PR #56, opened 2019-12-16
Filing: get qualityExists, et al. from /latest submission

### LINKED ISSUE #53: Filing: Fetch latest Submission from `/latest` endpoint for Institution component
## Goal
Save backend processing by removing `macroExists, macroVerified, qualityExists, qualityVerified` flags from the `/filing/institutions/{{lei}}/filings/{{year}}`

## Current behavior
In order to display the user's progress in the filing process, the Institution component currently uses the 4 mentioned flags from the latests submission from the above mentioned endpoint.  

## Expected behavior
Use the `/latest` submission endpoint to get the required flags instead of relying on  `/filing/institutions/{{lei}}/filings/{{year}}`

meissadia:
Closes #53 

## Changes

- Fetch latest Submission when fetching Filings for an Institution
- Create new Submission for an Institution when creating a new Filing
- New reducer, actions, and action types for `latestSubmissions`
- Fix: Submission router not loading due to missing `config` (seems to be working fine on dev-default but was failing locally... 🤷‍♂)

## Testing

1. I've tested with multiple Institutions associated via Keycloak 👍 
2. I've tested with new Institutions to verify that Filings and Submissions are created and retrieved as expected. 👌 
3. I've tested that the appropriate Institution is updated when uploading a new file / progressing through the filing process and that the Filing progress is accurately reflected on the `Institutions` page. :shipit: 

## Notes
- This was a tough one for me.  Happy to address other scenarios that should be tested to make sure I didn't break anything.


src/filing/actions/fetchLatestSubmission.js
```diff
@@ -0,0 +1,25 @@
+import hasHttpError from '../actions/hasHttpError'
+import receiveError from '../actions/receiveError'
+import receiveLatestSubmission from './receiveLatestSubmission'
+import requestLatestSubmission from '../actions/requestLatestSubmission'
+import { getNewestSubmission } from '../api/api'
```
wpears: `import { getLatestSubmission } from '../api/api'`

src/filing/actions/fetchLatestSubmission.js
```diff
@@ -0,0 +1,25 @@
...
+import { error } from '../utils/log.js'
+
+export default function fetchLatestSubmission(lei, filing) {
+  return dispatch => {
+    dispatch(requestLatestSubmission(lei))
+    return getNewestSubmission(lei, filing)
```
wpears: `return getLatestSubmission(lei, filing)`

src/filing/api/api.js
```diff
@@ -26,6 +26,10 @@ export function getLatestSubmission() {
   return fetchData({ submission: 'latest', noCache: 1 })
 }
 
+export function getNewestSubmission(lei, filing){
```
wpears: Remove this function and change `getLatestSubmission` to:

```
export function getLatestSubmission(lei, filing) {
  const options = { submission: 'latest', noCache: 1 }
  if(lei) options.lei = lei
  if(filing) options.filing = filing
  return fetchData(options)
}
```

src/filing/actions/fetchNewFiling.js
```diff
@@ -15,7 +16,10 @@ export default function fetchNewFiling(filing) {
             dispatch(receiveError(json))
             throw new Error(json && `${json.status}: ${json.statusText}`)
           }
-          return dispatch(receiveFiling(json))
+          if (!hasError) {
+            dispatch(receiveFiling(json))
+            return dispatch(receiveLatestSubmission({ lei: filing.lei }))
```
meissadia: I was creating a new submission after the creation of a new filing.

After discussing with Bhaarat, that would not accurately reflect the user's actions and also mess up some analytics. 

I'm currently unable to test my latest change because `dev-default` is unreachable, but I will now be adding a dummy submission to the `store` after creating a new filing.   I think this will allow the Institutions page to properly display the user's current filing progress without creating anything on the backend.

wpears: Ahh, yeah thanks for taking that out, I noticed that when creating my other branch but forgot to bring it up once I abandoned that approach

PR #73, opened 2019-12-19
Edits don't need to be fetched if code is >=14

### LINKED ISSUE #72: Correctly show navbutton when uploading a file with no edits

wpears:
Closes #72


meissadia:
Looks reasonable 👍

PR #93, opened 2019-12-26
Filing: avoid Summary fetch error

meissadia:
## Changes

- Reroute to the `/upload` page if the user is browsing to another page when the Submission status indicates that they should be on the `/upload` page.
- Skip rendering during the transition to the `/upload` page

## Testing

1. Complete and sign a filing
2. Click the `FILING HOME` link in the header nav
3. Click `Upload a new file`
4. Click 'Yes, replace HMDA data.`
5. Click the browser `Back` button
6. Observe that no error message is displayed at the top of the page

PR #94, opened 2019-12-27
Router cleanup and specifically render loader on submission page

### LINKED ISSUE #90: File upload page throws an error when file upload is cancelled in the middle

## Current behavior

Screen Shot 2019-12-23 at 9 43 34 AM

1.

wpears:
Closes #90 

Cleans up some router code and specifically blocks rendering router children on the submission page if the code isn't right.

Somewhat more idiomatic version of https://github.com/cfpb/hmda-frontend/pull/93 which will also prevent users from incorrectly getting to macro/submission page if they've unchecked the verification checkbox.

### DIFF
--- a/src/filing/submission/router.jsx
+++ b/src/filing/submission/router.jsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
 import { withRouter } from 'react-router-dom'
+import equal from 'fast-deep-equal'
 import SubmissionContainer from './container.jsx'
 import Loading from '../../common/LoadingIcon.jsx'
 import fetchSubmission from '../actions/fetchSubmission.js'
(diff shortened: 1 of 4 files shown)


src/filing/submission/router.jsx
```diff
@@ -97,37 +104,45 @@ export class SubmissionRouter extends Component {
...
 
   render() {
-    const { submission, match: {params} } = this.props
+    const { submission, lei, match: {params} } = this.props
+    const { code } = submission.status
+
     if (
-      submission.status.code === UNINITIALIZED ||
+      code === UNINITIALIZED ||
+      (code < VALIDATED && code !== NO_MACRO_EDITS && params.splat === 'submission') ||
```
meissadia: Just trying to understand what this scenario is...

User is trying to access the `signing` page but have not reached that part of the process.  What would be loading here?

PR #95, opened 2019-12-27
Add flagged maps/graphs to data-browser

wpears:
## Changes

- Adds maps/graphs code behind a `showMaps` flag in the config.

## Testing

1. Pull branch, note that maps/graphs are inaccessible (until merged, the flag doesn't exist so will be falsy)
2. Override the checks for `showMaps` in index.js/Home.jsx by prefixing them with `1 ||`.
3. The maps/graphs page should be accessible.

## Notes

- We could possibly add some local testing logic to the config setup to make the clunky testing above unnecessary when adding a flag, but I don't anticipate we're going to be adding a million.


src/data-browser/charts/MapContainer.jsx
```diff
@@ -0,0 +1,354 @@
...
+  const defaults = getDefaultsFromSearch(props)
+
+  const [map, setMap] = useState(null)
+  const [data, setData] = useState(null)
+  const [selectedVariable, setVariable] = useState(defaults.variable)
+  const [selectedValue, setValue] = useState(defaults.value)
+  const [fips, setFips] = useState(defaults.fips)
+
+  const fetchCSV = () => {
+    const csv = `/v2/data-browser-api/view/csv?years=2018&counties=${fips}&${selectedVariable.value}=${selectedValue.value}`
```
meissadia: When proxying against `dev-default` I am unable to download the CSV data.

When pointing to a locally running `data-browser-api` the downloaded data looks good.

meissadia:
I think the easiest way to locally test `config` changes is by commenting out 

https://github.com/cfpb/hmda-frontend/blob/master/src/common/useEnvironmentConfig.jsx#L8

src/data-browser/charts/MapContainer.jsx
wpears: This looks like a problem with `dev-default`, you'll notice that csv downloads are failing in the  dataset filtering page as well

wpears:
> I think the easiest way to locally test `config` changes is by commenting out
> 
> https://github.com/cfpb/hmda-frontend/blob/master/src/common/useEnvironmentConfig.jsx#L8

I pushed ba5d91f to make the default behavior more often what we want: in local dev, it will use your copies of config.json (since an update is just a `git fetch / merge` away) but will still grab and update state with the hosted version for all non-localhost environments.

Issue #103, opened 2020-01-02
Update deadline date in 2019 Institutions announcement 


meissadia:
## Current behavior
<img width="760" alt="Screen Shot 2020-01-01 at 11 14 47 PM" src="https://user-images.githubusercontent.com/2592907/71654155-67e2a780-2ced-11ea-83e4-026907149426.png">

## Expected behavior
Date should reflect the March 2nd deadline.

PR #106, opened 2020-01-02
Special case 2019 March 2nd filing deadline

wpears:
Closes https://github.com/cfpb/hmda-frontend/issues/103

### DIFF
--- a/src/filing/institutions/Header.jsx
+++ b/src/filing/institutions/Header.jsx
@@ -31,12 +31,15 @@ const InstitutionsHeader = ({ filingPeriod }) => {
       </Alert>
     )
   }
+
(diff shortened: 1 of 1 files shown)


src/filing/institutions/Header.jsx
```diff
@@ -31,12 +31,15 @@ const InstitutionsHeader = ({ filingPeriod }) => {
       </Alert>
     )
   }
+
+  const lastFilingDay = filingPeriod == '2019' ? '2nd' : '1st'
```
meissadia:
```suggestion
  const lastFilingDay = filingPeriod === '2019' ? '2nd' : '1st'
```

filingPeriod is coming through as a String so we may as well avoid the eslint warning about using `==`

PR #195, opened 2020-01-29
add enzyme and upgrade react-scripts

wpears:
Lays the groundwork for future testing work


wpears:
Should be ready to go, not exactly sure what's going on with that supposed CI failure

PR #246, opened 2020-02-13
Download links for Dynamic Datasets

### LINKED ISSUE #245: Make links to Dynamic Datasets "download"
2017 LAR is 1.6 GB
2018 LAR is 5.5 GB

It doesn't make sense to load these files in the browser.

meissadia:
Closes #245


wpears:
The cross browser solution here is to get the headers set appropriately in S3. I believe the bucket policy is supposed to set `Content-Disposition: attachment` but we've seen in the past that for some reason it isn't propagating to new files.

meissadia:
@wpears That's a good call.  I can open a DevOps ticket to update the headers instead.

wpears:
This has been implemented by bhargav, closing

PR #249, opened 2020-02-13
Add 'Instructions' to LARFT and FFVT doc sections

wpears:
Clarifies that the links are to instructions rather than the tools themselves


chynnakeys:
LGTM

PR #254, opened 2020-02-20
Note beta submission in receipt and remove irs in beta

### LINKED ISSUE #255: Remove IRS download link when running in beta

## Current behavior

IRS download link is always displayed even when the frontend is running in beta

## Expected behavior

We should remove the IRS download link when the frontend is running in beta. Additionally, we should clearly demonstrate when the filer signs on beta so that they know they aren't filing on the live platform.

1.

wpears:
Closes #255 
Helps users confused about whether they still need to file if they've submitted in beta


meissadia:
@wpears  Looks good locally.  Will merge once we have a link to the live Platform requested in #256.

PR #378, opened 2020-04-15
fix functioning of maintenance message by using plain state variable

### LINKED ISSUE #340: Maintenance banner immediately dismissed
Banner is automatically being dismissed on page load, at all resolutions.

Banner should only be closable on mobile and should only be dismissed on click of close button.  

maint-dismissed

wpears:
Closes #340


meissadia:
Couple issues:

- Using a plain variable doesn't trigger a refresh of the MaintenanceMessage component, hence why it is not dismissible on the Filing app page.  

- There's some padding that gets added to the footers when showing a maint. message, then (should be) removed when the message is dismissed.  Not sure this is happening for the Filing footer.

I've tried to tie it all together using shared state higher up in the component tree in [this branch](https://github.com/cfpb/hmda-frontend/compare/meis/fix-maintenance-message).

PR #413, opened 2020-04-28
Adjustable legend

### LINKED ISSUE #384: Add a legend to explain the color coding in the maps and graphs feature
Add a legend to the map displayed in the maps & graphs feature to clarify what numerical value each color represents.

### LINKED ISSUE #385: Specify data units for each variable represented in the maps and graphs feature
- Specify the data unit in the map legend. 
- Identify what units best represent each variable. 
- Allow the user to choose the unit level they would like to see represented (example: loan amount vs. loan count)

wpears:
Closes #384 
Closes #385

Instead of the base legend running between 0 - 32, this sets a bias based on the range of values present in each variable/value combination and adjusts the legend labels appropriately.

Screen Shot 2020-04-28 at 9 10 20 AM

### DIFF
--- a/src/data-browser/charts/MapContainer.jsx
+++ b/src/data-browser/charts/MapContainer.jsx
@@ -29,7 +29,11 @@ const legendBody = colors.map((color, i) => {
   return (
     <div className="legWrap" key={i}>
       <span className="legColor" style={{backgroundColor: color}}></span>
-      <span className="legSpan">{i*4} - {(i+1)*4}</span>
+      <span className="legSpan">{
+        i  === colors.length -1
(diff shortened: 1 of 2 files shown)


meissadia:
@wpears This looks awesome!  I think the color variation is a smart trigger for users to look at the legend to understand why it's different. 

I think the one bit of feedback from your presentation that still needs to be addressed was that the ranges (0 - 1, 1 - 2) should not overlap (0 - 1, 2 - 3).  

Also super randomly noticed that 02158 - Kusilvak Census Area, AK is missing it's name
Screen Shot 2020-04-30 at 12 18 59 PM

PR #427, opened 2020-05-07
[DB] Fix react-select drop-down in IE/Chrome

### LINKED ISSUE #421: Data Browser - Inconsistent behavior using the vertical scroll in the drop downs
Inconsistent behavior using the vertical scroll bar:
In IE:
- Sometimes the drop down box disappears when trying to scroll
- The up down arrows also don't seem to move the list

In Chrome:
- Scrolling through the dropdown seems to be fine but when using the arrow keys the blue highlight goes down, but the location does not so all we see are the top entries in the list

meissadia:
Closes #421 

Tested locally using [IE Tab](https://chrome.google.com/webstore/detail/ie-tab/hehijbfgiekmjfkfjpbkbammjbdenadd/related?hl=en-US).

- [x] Add keyboard (up/down, page-up/down) navigation to drop-downs (IE/Chrome)
  - Changes the MenuList component to scroll to the focused element
  - Works in Chrome, does not work in IE11
- [x] Address vertical scrollbar disappearing on click (IE)
   - Not sure of the root cause.  There was some indication in the GH issues for react-select that upgrading the version could provide resolution, which is included in this PR. 
  - IE Tab testing is showing that this is still an issue.


src/data-browser/geo/MenuList.jsx
```diff
@@ -3,18 +3,46 @@ import { FixedSizeList as List } from 'react-window'
...
+  constructor(props) {
+    super(props)
+    this.listRef = undefined
+  }
+
+  getFocusIndex = (children) => {
+    const isArray = children instanceof Array
+    children = isArray ? children : [children]
+    return Math.max(
+      children.findIndex(({ props: { isFocused } = {} } = {}) => {
```
wpears, line 16: findIndex will have to be replaced with a small for loop, as it's unsupported in ie

PR #441, opened 2020-05-19
update cypress dockerfile to use centos7 as base image

### LINKED ISSUE #463: update cypress dockerfile to use centos7

BarakStout:
Change Cypress to use Centos7 as the base image and upgrade Cypress to 4.6.0

Closes #463


.yarnrc
```diff
@@ -6,4 +6,4 @@ lastUpdateCheck 1496303126136
 yarn-offline-mirror "./npm-packages-offline-cache"
 yarn-offline-mirror-pruning true
 --install.pure-lockfile true
---install.offline true
+#--install.offline true
```
meissadia, line 9: I don't think we want this permanently commented.

I've got the image building by doing the follow:

.yarnrc
```
#--install.pure-lockfile true
#--install.offline true
```

`yarn cache clean`
`yarn add cypress@4.6.0`

.yarnrc
```
--install.pure-lockfile true
--install.offline true
```

`docker build -t cypress-bs-01 -f cypress/Dockerfile .`
`docker run -it cypress-bs-01`

Able to run the Cypress Tools tests after setting the appropriate ENV vars.

BarakStout: Updated

PR #450, opened 2020-05-20
State geography selector

wpears:
## Changes

- Factors out a decent bit of code into separate modules
- Adds a new geography selector and updates code to respond to different geographies

## TODO
  - Highlighting and tooltip performance is a bit wonky
  - Handlers need to be fixed to support multiple geographies


meissadia:
@wpears I think you've already got these covered in your TODOs but just to note:
- Switching from State => County
  - selected State outline persists
- State Geography
  - After selecting a State, should a summary table be displayed?
  - Landmarks (city names) covered
    - Is this an issue?
      - Helps orient when zoomed in but also looks kind of nice without the additional visual noise.

PR #549, opened 2020-07-06
Update Cypress Quarterly tests

### LINKED ISSUE #544: Update Cypress Quarterly tests
- [x] Use config to identify open/close periods
- [x] Verify that we can't file for close quarters
- [x] Add 2020-Q2 test & LAR fixture

meissadia:
Closes #544 

- Add Q2 and Q3 test LAR files
- Use config to determine if period is open/closed and how to test


src/common/configUtils.js
```diff
@@ -26,3 +24,7 @@ export function getDefaultConfig(host) {
 export function isProd(host) {
   return !!host.match('^ffiec')
 }
+
+export function isBeta(host) {
```
meissadia, line 28: Cypress was having an issue with the syntax of the CSS file imported in the `Beta` component.  I could not figure out why in the world that would be the case, so I had to just redefine the `isBeta` function.

PR #579, opened 2020-07-24
Fix DB Tests

### LINKED ISSUE #577: Fix Data Browser tests
Still seeing intermittent failures in the DB tests due to point/click actions selecting the wrong thing or not being able to select anything.  

Need to stabilize these.

meissadia:
Closes #577 

The Institution selector tends to steal the page focus when the populating API call completes.  This was causing Cypress' attempts to enter values in other inputs to be prematurely truncated, leading to inconsistent app state and intermittent failures.  

This PR makes it so that the Cypress tests wait until the initial Institution selector load completes before starting to enter values. 

Also changes from a simulated point-and-click selection to typing values into the input fields to provide more consistent results on what was being chosen from the selection inputs.


meissadia:
CI tests on travis-ci.org are failing due to missing environment configuration, which I don't have access to modify.  
[continuous-integration/travis-ci/pr](https://travis-ci.org/github/cfpb/hmda-frontend/builds/711835554)
[continuous-integration/travis-ci/push](https://travis-ci.org/github/cfpb/hmda-frontend/builds/711835552)

Those tests are redundant and I've opened https://github.com/cfpb/hmda-frontend/issues/580 to stop them from triggering, which I also don't have access to change.

PR #649, opened 2020-09-02
View Institution Details

### LINKED ISSUE #636: Provide ability to view institution record

## Current behavior
There isn't a way for users to view their own institution details

## Expected behavior
Call the GET institution endpoint so that institutions details can be shown to the users. This will help institutions resolve S303 error which requires the agency code and the taxid submitted in the transmittal sheet to match the institution details on our end. 

1.

meissadia:
Closes #636 

## Changes
- Adds Bank 0 to CI test Institution generation to make it easier to work with Platform test files
- Adds an icon on the Institutions overview page to link to the Institution Details view
- Displays most frequently used Institution Details by default
- Displays additional fields as requested
- Stores the newly required field data from the Institutions endpoint in Redux to avoid the need to refetch

## Screenshots
Button
  button

Default Fields
  min

All Fields
  more


src/filing/institutions/container.jsx
```diff
@@ -34,8 +43,28 @@ export class InstitutionContainer extends Component {
     }
   }
 
+  updateUrlIfNeeded(){
```
wpears, line 46: I think that instead of managing the url ad hoc here, that this should be incorporated into the route for the institution page directly. Namely, I think an `/:institution` argument should be added [in the index file](https://github.com/cfpb/hmda-frontend/blob/master/src/filing/index.js#L62) which then can be passed down and used by the institution container to control routing. The info button just becomes a `Link` to `/filing/<year>/institutions/<institution/` ( I think we can drop `/details` since it's kinda self-evident). And the back button from the details page is just `Link` back to `/filing/<year>/institutions/`

src/filing/institutions/details/InstitutionDetailsWrapper.jsx
```diff
@@ -0,0 +1,39 @@
...
+    if (!window.location.pathname.match(/details$/)) onClose()
+    // eslint-disable-next-line
+  }, [window.location.key, match])
+
+  if (!institutions || !institutions.fetched) return wrapLoading()
+  institutions = institutions.institutions
+
+  return (
+    <main id='main-content' className='institutions-details full-width'>
+      <button className='back' type='button' onClick={onClose}>
```
wpears, line 31: I think this button should have the `button` class to match the other buttons (then you can add other rules to the `back` class to control positioning).

wpears (changes requested):
Looks great overall, one design and one structural comment.

src/filing/institutions/container.jsx
meissadia: Much cleaner and more sensible than the fiasco I pieced together.  🤦🏽  
 
Thank you!  🚀

BarakStout (approved):
LGTM 👍

wpears (approved):
Nice work!

PR #705, opened 2020-09-28
document generate_data.js script

wpears:
Adds usage info + docs for `generate_data.js`


wpears:
Added in internal repo, merging here for now as well (we may want to remove this if they look like they'll get out of sync)

PR #716, opened 2020-10-08
648 publication changes

### LINKED ISSUE #648: Publications Change Log
New datasets and products are being released and prior datasets are receiving material updates. These updates, changes, and new products need to be communicated to end users. Creating a section that contains information similar to patch notes would provide us an area to publish metadata regarding updates.

## Current behavior
No updates are supplied to data users regarding changes to datasets or data products.

## Expected behavior
Users are made aware of changes made to datasets and data products so they can assess impacts to their work.

meissadia:
- Adds new Publication Updates page
- Links to Updates page from Publications

### Link to Publication Updates page
Screen Shot 2020-10-08 at 1 47 45 AM

### Updates Page
pub-changes

### DIFF
--- a/src/data-publication/PublicationChanges.jsx
+++ b/src/data-publication/PublicationChanges.jsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from 'react'
+import { Link } from 'react-router-dom'
 import LoadingIcon from '../common/LoadingIcon'
 import defaultData from './constants/publication-changes.json'
 import './PublicationChanges.css'
@@ -55,7 +56,8 @@ const PublicationChanges = () => {
   if (loading) return <LoadingState />
 
   return (
-    <div id='publication-changes' class='full-width'>
+    <div id='publication-changes' className='full-width'>
+      <div className='BackLink'><Link to='/data-publication/'>{'\u2b05'} Back to HMDA Data Publications</Link></div>
       <h1>Data Publication: Updates and Notes</h1>
       <p className='intro'>
         The HMDA data and reports are the most comprehensive publicly available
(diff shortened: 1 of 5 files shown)


src/data-publication/PublicationChanges.jsx
```diff
@@ -0,0 +1,141 @@
...
+const PRODUCT_KEYS = ['mlar', 'datasets', 'aggregate', 'disclosure']
+
+const PRODUCT_TITLES = {
+  mlar: 'Modified LAR',
+  datasets: 'National Loan Level Datasets',
+  aggregate: 'Aggregate Reports',
+  disclosure: 'Disclosure Reports',
+}
+
+const FILE_URL = 'https://raw.githubusercontent.com/cfpb/hmda-frontend/master/src/data-publication/constants/publication-changes.js'
```
wpears, line 21: Should be `.json` afaict

wpears (changes requested):
I wonder if we should wrap the data fetching similar to how we do for the environment config? I think it makes sense to generalize the default to what's built -> look for fresher data behavior in a way that is easier to drop in and won'd require all the fetch/load managing logic in each component that needs this behavior.

meissadia:
> I wonder if we should wrap the data fetching similar to how we do for the environment config? I think it makes sense to generalize the default to what's built -> look for fresher data behavior in a way that is easier to drop in and won'd require all the fetch/load managing logic in each component that needs this behavior.

Addressed on branch `648-alt-pub-changes`: [useRemoteJSON](https://github.com/cfpb/hmda-frontend/blob/e685fe37dacbaf89cd76c1bd8665a6796fa1ff99/src/common/useRemoteJSON.jsx)

PR #718, opened 2020-10-08
[Cypress] Pull correct env config file

### LINKED ISSUE #717: Cypress failing Quarterly Filing tests in Prod-Beta
This was the result of a couple issues:
- [x] FRONTENDTESTBANK9999 use was not set as a `Quarterly Filer`.  This has been updated via HMDA Help and now the Quarterly tests successfully complete.
- [x] Cypress is pulling the wrong config file (dev-beta instead of prod-beta) and thus was not running `2020-Q3`.

meissadia:
Closes #717 

Strips http:// and https:// from host name when determining environment configuration file.

### DIFF
--- a/src/common/configUtils.js
+++ b/src/common/configUtils.js
@@ -11,7 +11,8 @@ export function fetchEnvConfig(setFn, host) {
     .then(config => setFn(config))
 }
 
-export function getDefaultConfig(host) {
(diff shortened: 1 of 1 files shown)


BarakStout (approved):
LGTM 👍

PR #760, opened 2020-11-23
Upgrade to Cypress 6 and address Cypress HMDA Help failures

### LINKED ISSUE #758: Fix Cypress failures
Our upgrade to Cypress 5 has exposed (or created?) some flake in our HMDA Help test suite, particularly when attempting to log in.  

Because this error occurs in the `beforeEach` block, the whole test fails.  

- [x] Try moving login/logout behavior into the test block, this might allow the test to retry (without the new retry config)?
- [x] It seems like, if we leave the tests structured as is, the new retry configuration would still fail in this instance because of Cypress' behavior when there's an error in the `beforeEach` hook. But try the config-only route anyway.
- [x] These tests don't run on CI so I'll need to verify locally, but I don't see this failure as consistently as the deployed test suite does, so that might slow down the progress.

meissadia:
Closes #758 

- Uses Cypress 6 and adds retries (default 3x in CI)
- Moves auth actions into the test block to support retry on login failure.
- Adds Cypress function `#logEnv` to log environment variables in the Cypress logs for easier debugging via video review.  
- Leverage a single User for all test requiring authentication (i.e. Filing, Help) to simplify configuration.


BarakStout (approved):
LGTM 👍

PR #765, opened 2020-11-24
Update popup styling to be more clear

wpears:
The FIPS isn't really helpful in the popup and (especially for states) could be confusing. The `per 1000 people` bit I think makes it less of a floating number with no context


meissadia (approved):
I like it 👍🏽

PR #802, opened 2020-12-21
Use defaultDocsPeriod instead of defaultPeriod to set period for Tools

### LINKED ISSUE #788: Update Documentation and Tools to default to 2020
- [x] [LARFT](https://ffiec.cfpb.gov/tools/lar-formatting) doc link
- [x] [FFVT](https://ffiec.cfpb.gov/tools/file-format-verification) tool default

meissadia:
Closes #788 again.

Uses `defaultDocsPeriod` instead of `defaultPeriod` so that the original intent is met: to allow Docs/Tools to point to the upcoming Filing season separately from the Filing app options.

### DIFF
--- a/src/tools/file-format-verification/reducers/index.js
+++ b/src/tools/file-format-verification/reducers/index.js
@@ -40,7 +40,7 @@ const defaultPagination = {
   fade: 0
 }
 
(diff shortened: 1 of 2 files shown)


BarakStout (approved):
LGTM 👍

PR #839, opened 2021-01-29
Display HTTP Status Code in Filing errors

### LINKED ISSUE #776: Include HTTP status code in Filing error message

## Current behavior
If a 504 error is thrown by the network in middle of the file upload process, the frontend shows "Sorry an error has occurred" and halts the upload process. However, if the users refresh their page after a while, the file upload has actually finished. 

## Expected behavior
This should work as is if the issues with load balancer are resolved. However, in the meantime we might be able to do the below:

If during the file upload process a HTTP 504 is thrown, don't show "Sorry an error has occurred" error to user. Rather just suppress the error and continue. 

1.

meissadia:
Close #776 

- ~Suppresses 504 errors that occur during file upload.~
- Displays status code and status text when available
  Screen Shot 2021-01-29 at 2 57 43 PM

### DIFF
--- a/src/filing/common/ErrorWarning.css
+++ b/src/filing/common/ErrorWarning.css
@@ -1,3 +1,9 @@
 .ErrorWarning {
   margin: 2em 0;
 }
+
+.ErrorWarning .error-code{
+  display: block;
+  margin-top: .5rem;
+  font-weight: bold;
+}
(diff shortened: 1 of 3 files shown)


PatrickGoRaft (approved):
LGTM.

PR #851, opened 2021-02-20
[Help] Download Submission file

### LINKED ISSUE #698: UI to Download Submission File
- [x] Endpoint
- [x] Find good place in UI for this download link

meissadia:
Closes #698 
### Changes
- Refactors to share common Keycloak / Auth Token code.
- Add download of Submission files via the StreamSaver library to bypass the the `Blob` 500mb limit. 
### Review Focus
- Any concerns about [service worker or MITM](https://github.com/cfpb/hmda-frontend/pull/851/commits/ad92868e8e79f17bdc81386877c98b8428e823aa)?


wpears (commented):
The main trouble with the ServiceWorker approach is that it won't be usable on IE11, which I believe is a significant portion of filers (this can be checked, but at one point IE11 commanded a plurality of the share of browsers used by filers). As far as the implementation goes, it's a pretty interesting system that is a nice hack around the Blob limit, but due to the IE11 limitation I think another approach may need to be tried (ie, the backend returning the response as a file instead of as a data endpoint).

meissadia:
@wpears The audience for this functionality is the [HMDA Help team](https://github.com/cfpb/hmda-frontend/pull/851/commits/376e3f434755d62b14d978efa2b834374914f1cf) rather than Filers, so while it may not be ideal, we do have some leeway in asking them to use a non-IE browser when they need to access this feature.

wpears:
Oh, of course! Yeah, then I'd say this strategy should work out decently. I'm a little worried about this being finnicky/brittle, but if it's an internal-only audience I think that that's an acceptable risk.

BarakStout (approved):
LGTM 👍

Issue #852, opened 2021-02-23
Update Rate Spread error message for 7/21/20


ericspry:
Update the Rate Spread calculator page to replace the current error message for 7/20/20 with updated message.

**Remove the following:** 7/21/2020: An error in the Mortgage Rate Survey table for 15-Year Fixed Points & Fees (cell E135) resulted in incorrect results for Rate Spread on fixed rate loans with terms of 13 to 22 years and rate set date of 7/20/2020. The issue was corrected in the online calculator and the linked tables.

**Replace with:** 02/23/2021: Two sets of APORs were published for the week of 7/20/2020 for fixed rate loans with terms of 13 to 22 years. The first set was published on 7/16/2020 and was briefly incorporated into the Bureau’s rate spread calculator. The second set was published on 7/20/2020 and was substituted for the first in the Bureau’s rate spread calculator. Both sets of APORs are available here [link to table]. 

I'll share the csv

PR #854, opened 2021-02-23
Update rate-spread error message (and fix keycloak mock function)

wpears:
Closes https://github.com/cfpb/hmda-frontend/issues/852


ericspry:
I think first use of files.cf for HMDA Platform files.. but I think I prefer that. LGTM

meissadia (approved):
Noice. Thanks, @wpears!

PR #919, opened 2021-04-16
[DynamicDatasets] Remove Panel documentation links

### LINKED ISSUE #916: Remove Panel Documentation from Dynamic Page

## Current behavior
image

There isn't a dynamic panel dataset, so the file specs on the right side ("Public Panel Schema, Public Panel Values and Definition") don't need to be linked on this page. They only need to be linked on the Snapshot page.

meissadia:
Closes #916

- Removes extraneous file spec links for Panel
- Rewords and reorders file spec links for LAR files to better connect the two documents
Screen Shot 2021-04-19 at 4 51 32 PM


BarakStout (approved):
LGTM 👍

PR #931, opened 2021-04-20
[Help] New Institution search

### LINKED ISSUE #918: HMDA Help: Institution search box suggestions

meissadia:
Closes #918

### Changes
- Search by LEI or Institution name
- Auto-search on selection
- Input sanitization to handle spaces and non-alphanumeric characters that get pasted in
- Additional indicators when user appears to be inputting an LEI vs searching for a name
- Updated Cypress tests

### Demo
new-help-search

### DIFF
--- a/cypress/integration/data-browser/2018.spec.js
+++ b/cypress/integration/data-browser/2018.spec.js
@@ -4,7 +4,7 @@ const { HOST, ENVIRONMENT } = Cypress.env()
 const dbUrl = dbURL.bind(null, HOST)
 
 describe('Data Browser 2018', function () {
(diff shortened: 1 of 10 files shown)


wpears (approved):
This is great!

My main comment is on the flow/signaling around creating a new institution. I don't think there should be a `CREATE "BLAH"` entry in the list unless the LEI is 20 characters. Maybe instead a note saying "LEIs are 20 characters" or something like that. Then once it's 20 characters switch it to the CREATE message.

That could/should probably be done in a separate PR :)

✅  for this first push

PR #979, opened 2021-05-21
fix(maps): improve layout for mobile views

### LINKED ISSUE #978: Maps: Better mobile views
UI does not degrade gracefully on smaller screens.

meissadia:
Closes #978
- Views tested via browser dev tools.
- Opens Maps in Beta, which can be accessed from my mobile device and thus used for live testing of these changes. Data Browser is not presented to users on the Beta homepage.
 
## Before
 before

## After
after

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -16,5 +16,6 @@
   "retries": {
     "runMode": 1,
     "openMode": 0
-  }
+  },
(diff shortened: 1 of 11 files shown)


BarakStout (approved):
LGTM 👍

PR #997, opened 2021-06-03
fix(maps): App crashes when clearing only selected filter

### LINKED ISSUE #998: Maps crash
This should have been fixed in v1.3.4 but I must have forgotten to push the commit, because the bug persisted. 

issue: After selecting a filter AND a feature, clearing the filter selection would result in a crash of the app.  

fix: guard against the empty selection

meissadia:
Closes #998


BarakStout (approved):
LGTM 👍

PR #1042, opened 2021-07-13
Data Browser - Select All button

### LINKED ISSUE #114: Add 'Select All' checkbox
Include a ‘Select All’ variable checkbox that auto selects all options so users don’t have to select each sub-variable individually.

Note: Add this to the backlog based on feedback from initial testing; we should check to see if this is still a usability issue during future testing. If not, we should consider removing this.

meissadia:
Closes #114 

Adds a button to each Filter option selector in the Data Browser to either "Select All" or 
Clear Selected".

db-select-all-proto-3

### DIFF
--- a/src/data-browser/constants/variables.js
+++ b/src/data-browser/constants/variables.js
@@ -204,7 +204,8 @@ function makeObj(label, definition) {
 
 function buildWithId(label, definition, list) {
   const obj = makeObj(label, definition)
-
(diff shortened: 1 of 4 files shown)


meissadia:
@BarakStout This has been updated with the styling feedback from Standup.

BarakStout (approved):
LGMT 👍

PR #1049, opened 2021-07-26
fix(databrowser): Remove option ID from all displayed instances

### LINKED ISSUE #1047: Update Cypress tests
Tests need to be updated to eliminate false-negatives in order to restore confidence in the monitoring jobs.

meissadia:
Closes #1047 

In the HMDA Data Browser, we previously removed the numeric Option ID from the checkbox label, but did not make the change consistent across the other page elements that display the label. This PR corrects that inconsistency in both the UI and the Cypress tests.

|before|after|
|--|--|
|Screen Shot 2021-07-26 at 2 40 43 PM|Screen Shot 2021-07-26 at 2 41 00 PM|


BarakStout (approved):
LGTM 👍

PR #1089, opened 2021-08-26
1 - Begin extended downtime. Filing app closed for maintenance.

meissadia:
- Prevents user access to the Filing application
- Displays notification banner of maintenance activities


PatrickGoRaft (approved):
LGTM.

PR #1097, opened 2021-08-27
Revert upgrade of @testing-library/cypress to 8.0.0

### LINKED ISSUE #1096: Revert @testing-library/cypress to 7.0.3
Containerized Frontend uses node 14.9.0, while my dev environment uses 12.14.1.  As such, my local testing did not catch an incompatibility with node when attempting to upgrade @testing-library/cypress to v8.0.0.
 
When building the image, I see a fatal error due to this incompatablity.

Plan:
- Revert @testing-library/cypress
- Add note to upgrade this library when we upgrade node/nginx in #942

meissadia:
Closes #1096


PatrickGoRaft (approved):
LGTM.

PR #1126, opened 2021-09-23
updates to cypress docker build

### LINKED ISSUE #381: Run cypress as non root user
After #380 is merged in, we should update `cypress/Dockerfile` to run as non root

BarakStout:
Closes #381 

- Updates Cypress dockerfile to use 7.4.0 
- Updates dockerfile to a non-root user


meissadia (approved):
Nice

PR #1129, opened 2021-09-27
HMDA Mailing List

### LINKED ISSUE #1007: Add a link on the FFIEC website to the Update Signup Page
The website https://content.consumerfinance.gov/compliance/compliance-resources/signup/ is public & allows users to signup for updates regarding HMDA data. The HMDA section is not visible yet, but will be soon. Creating a ticket to discuss where we would link to this on the HMDA website.

meissadia:
Closes #1007 

## Filing App Homepage
Screen Shot 2021-10-14 at 2 58 08 PM

Screen Shot 2021-10-14 at 2 57 57 PM

Screen Shot 2021-10-14 at 3 58 03 PM

## Site Footer
Screen Shot 2021-09-27 at 3 44 58 PM

Screen Shot 2021-10-14 at 3 05 12 PM

Screen Shot 2021-10-14 at 3 04 57 PM

Screen Shot 2021-10-14 at 3 58 17 PM


src/App.jsx
```diff
@@ -30,12 +30,12 @@ const HmdaHelp = makeAsyncComponent(() => import('./hmda-help'),
 const App = () => {
   const isFiling = !!window.location.pathname.match(/^\/filing/)
   const isHelp = !!window.location.pathname.match(/^\/hmda-help/)
-  const hideHeaderFooter = isFiling || isHelp
+  const hideCommonHeader = isFiling || isHelp
```
lchen-2101, line 33: maybe this can become `showCommonHeader`, then below, u can do `{showCommonHeader && <Rout ...` instead of having to do `null` later

src/common/mailingListServices.jsx
```diff
@@ -0,0 +1,145 @@
+import React from "react"
+import { useState } from "react"
```
lchen-2101, line 2: quotes, also, u should be able to combine these 2 into 1, can't u? `import React, { useState } from 'react'`

src/common/mailingListServices.jsx
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * Wrapper for displaying status alert
+   * @param {string} type Determines the Alert color scheme
+   * @param {string} message Text content
+   */
+  const Status = ({ type, message }) => {
+    if (!message) return
+    return (
+      <Alert type={type}>
+        <>{message}</>
```
lchen-2101, line 32: don't think we need fragments here do we? since it's wrapped inside `Alert` as parent already.

src/common/mailingListServices.jsx
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    request.open("POST", url)
+    request.setRequestHeader(
+      "Content-Type",
+      "application/x-www-form-urlencoded"
+    )
+    request.setRequestHeader("X-Requested-With", "XMLHttpRequest") // Required for cf.gov to send JSON response
+    request.send(content)
+  }
```
lchen-2101, line 98: we might want to think about using a common xhr / rest library instead of manually creating XHRs now; `Axios` is a pretty popular one, I'm sure there are others as well.

src/common/mailingListServices.jsx
```diff
@@ -0,0 +1,145 @@
...
+    error,
+    setError,
+    processing,
+    setProcessing,
+    invalid,
+    setInvalid,
+    onEmailChange,
+    onSubmit,
+    currentStatus,
+    submitButton,
```
lchen-2101, line 143: having components exported as part of the custom hook feels a bit awkward; also, this is a lot of exports... aside from the emailAddress, should the component using this be able to set the invalid, processing, success, error states?

lchen-2101 (changes requested):

src/common/MailingListSignup.jsx
```diff
@@ -0,0 +1,86 @@
+import React from "react"
```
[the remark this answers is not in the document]
meissadia: After searching around for quoting standards, there doesn't seem to be a ton of consensus on which is correct.  From some quick testing, even HTML attributes written with single quotes in JSX are output using double quotes in the rendered HTML. I've updated my Prettier settings to prefer single quotes and will use that configuration going forward.

src/common/MailingListSignup.jsx
```diff
@@ -0,0 +1,86 @@
...
+  ? "https://www.consumerfinance.gov"
+  : `${process.env.REACT_APP_CFGOV_DEV_DOMAIN}`
+
+// CF.gov endpoint for subscribing to a topic
+const SUBSCRIPTION_ENDPOINT = `${CFGOV_HOST}/subscriptions/new/`
+
+// HMDA topic for Production/Staging
+const HMDA_FILING_TOPIC_ID = isProd(window.location.host)
+  ? "USCFPB_139"
+  : "USCFPB_51"
```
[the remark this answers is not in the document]
meissadia: 👍🏽

src/common/mailingListServices.jsx
meissadia: Agreed, that was a lot of exports.  I've reduced the list to those elements that are actually imported by our Signup components. 
 
As for having components exported from a hook, it feels like a reasonable solution for this use case.  We've got a couple of Signup forms that need to display the same components in different layouts.  Those shared components rely on, and display, the current state stored in the `useSubscriptionLogic` hook.  Having those shared components be pre-composed means we don't need to duplicate that composition logic in two places.

src/common/mailingListServices.jsx
meissadia: Agreed, this is an awkward and verbose way of handling these requests.  I think integrating Axios should be spun off as it's own ticket since there are multiple places we'd want to replace these calls.  Thanks for calling this out 👍🏽

lchen-2101 (approved):
LGTM :shipit:

PR #1136, opened 2021-10-01
Update Cypress testing pod build process

### LINKED ISSUE #1135: Cypress automated testing

meissadia:
Closes #1135 

- Expand definition of compatible Node engines
- Update Cypress installation method in the testing pod

### DIFF
--- a/cypress/Dockerfile
+++ b/cypress/Dockerfile
@@ -33,16 +33,9 @@ RUN yum install -y unzip
 
 RUN npm install yarn -g
 
-RUN mkdir -p ~/.cache/Cypress/8.3.0/Cypress && \
-    wget --no-check-certificate http://download.cypress.io/desktop/8.3.0?platform=linux -O /tmp/z.$$ && \
-    unzip -d ~/.cache/Cypress/8.3.0/ /tmp/z.$$ && \
-    rm /tmp/z.$$ && \
-    chmod +x ~/.cache/Cypress/8.3.0/Cypress/Cypress && \
-    rm -rf /var/lib/apt/lists/*
-
 COPY / hmda-frontend/
 
-WORKDIR hmda-frontend/
+WORKDIR /hmda-frontend/
 
 RUN yarn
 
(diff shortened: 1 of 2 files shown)


cypress/Dockerfile
```diff
@@ -33,20 +33,15 @@ RUN yum install -y unzip
...
 RUN npm install yarn -g
 
-RUN mkdir -p ~/.cache/Cypress/8.3.0/Cypress && \
-    wget --no-check-certificate http://download.cypress.io/desktop/8.3.0?platform=linux -O /tmp/z.$$ && \
-    unzip -d ~/.cache/Cypress/8.3.0/ /tmp/z.$$ && \
-    rm /tmp/z.$$ && \
-    chmod +x ~/.cache/Cypress/8.3.0/Cypress/Cypress && \
-    rm -rf /var/lib/apt/lists/*
-
 COPY / hmda-frontend/
```
lchen-2101, line 36: should this line be changed to absolute path as well?

cypress/Dockerfile
```diff
@@ -33,20 +33,15 @@ RUN yum install -y unzip
...
+WORKDIR /hmda-frontend/
 
 RUN yarn
 
 RUN useradd -ms /bin/bash hmda_cypress_user && \
   chown -R hmda_cypress_user:hmda_cypress_user /hmda-frontend/
 
 USER hmda_cypress_user
+
+RUN yarn cypress install
```
lchen-2101, line 47: nice, I'm guessing this replaces the need to do all the downloading explicitly done by the block you removed above?

cypress/Dockerfile
meissadia: I adjusted the `WORKDIR` path because my IDE was giving me a warning, something to the effect that `WORKDIR paths should be absolute`.  Since I didn't get a similar warning for the `COPY` command, I left it as is. That said, things are working as currently configured, so I think we can leave this as relative.

cypress/Dockerfile
meissadia: Yes. The new command handles the download/installation to appropriate directory for the `hmda_cypress_user`. With the previous install method, the Cypress binary seemed to be getting overwritten, which might just have been an order-of-operations issue, but I think having this new command provides a much more succinct, understandable, and maintainable approach.

lchen-2101 (approved):
LGTM

PR #1159, opened 2021-10-18
ChangeLog - Refactor tests for easier readability

### LINKED ISSUE #1157: [Cypress] More dynamic test criteria for Change Log
When testing the Change Log text filter, we currently check for a specific number of output results, which is fragile since the number will change as new entries are created.

meissadia:
Closes #1157 

- Split tests into smaller chunks
- Reduce the number of static, numeric checks

## Testing
Run ChangeLog.spec.js against localhost or Production

Screen Shot 2021-10-18 at 3 27 38 PM


meissadia:
> did the js quotes standard change? I've always thought single quotes are the common standard within js.

I've been letting my code editor use it's default.  Updated the PR to use single quotes. We'll ensure this is enforced when we get around to adding a linting config.

lchen-2101 (approved):
LGTM

PR #1187, opened 2021-11-04
[Submission] Edits - Configurable per-Edit transformations

### LINKED ISSUE #859: Uniquely display Q656 & 657 data fields that trigger the edit

## Current behavior
Currently, q656 and 657 can potentially be triggered by at least 15+ fields. Given this, the table that displays the edit is extensive. 

## Expected behavior
We should find a way to uniquely display the specific data fields that trigger these edits. Some options may be highlighting, changing the order, only displaying the data fields that trigger the edits.

## Screenshots
image

meissadia:
Closes #859 

For Edits Q656 & Q657
- Displays only fields for which _some_ LAR contains the value `1111`
- Highlights fields of a LAR row that contains the value `1111`

Highlight-1111


meissadia:
## Paginated

I think these screenshots highlight how this approach can help users quickly identify problematic fields.  If only one field on a page is affected, only that field is shown.  Otherwise, fields of a row which are affected are highlighted, while non-affected fields are displayed in a plain style.

|Page 1|Page 2|
|---|---|
|Screen Shot 2021-11-04 at 4 22 55 PM|Screen Shot 2021-11-04 at 4 23 05 PM|

src/filing/actions/receiveEdit.js
```diff
@@ -1,14 +1,66 @@
...
+    r.fields = r.fields.filter((f) => {
+      if (!fieldMatches[f.name]) return null
+      if (f.value.match(pattern)) {
+        // Highlight the offending field
+        f.value = <span className='highlight'>{f.value}</span>
+      }
+      return f
+    })
+  })
+}
```
lchen-2101, line 33: I think generally we'd like to avoid side-effect style of coding if we can; so rather than modifying the argument passed in, I think returning a result would be cleaner; also I believe 1 iteration should be able to achieve this, instead of looping through the collection twice; think we can do `rows.map(...)` and return that result.

src/filing/actions/receiveEdit.js
```diff
@@ -1,14 +1,66 @@
...
+    type: 'regex',
+    filter: filterEditFields,
+    pattern: PATTERN_1111,
+  },
+}
+
 export default function receiveEdit(data) {
+  const useSpecial = TRANSFORMERS[data.edit]
+  if (useSpecial && useSpecial.type === 'regex')
+    useSpecial.filter(data.rows, useSpecial.pattern)
```
lchen-2101, line 54: keeping with the no side-effect theme, we can do something like `const rows = shouldUseSpecial ? filter(data.rows, ...) : data.rows`

src/filing/actions/receiveEdit.js
```diff
@@ -1,14 +1,66 @@
...
+
   return {
     type: types.RECEIVE_EDIT,
     edit: data.edit,
     rows: data.rows,
```
lchen-2101, line 59: then this can just be `..., edit: data.edit, rows, pagination: ...`

src/filing/actions/receiveEdit.js
```diff
@@ -1,14 +1,66 @@
+import React from 'react'
```
lchen-2101, line 1: I think from a certain version of react and onwards this is no longer needed; I'm guessing our version still pre-dates that, but just want to call it out here for now.

src/filing/actions/receiveEdit.js
```diff
@@ -1,14 +1,66 @@
...
 import * as types from '../constants'
 
+const PATTERN_1111 = /1111/
```
lchen-2101, line 4: do we need to worry about number formatting for this, or is it always going to be just 4 `1`s in a row? i.e. `1,111`, `1,111.00`, etc; also this pattern would capture anything with `1111` in it as well, is that desired? i.e. `121111324 a street`, if we only care about the sequence of 4 `1`s, we could do `/^1111$/

lchen-2101 (changes requested):

src/filing/actions/receiveEdit.js
meissadia: Tried removing it but got an error.  :+1:

src/filing/actions/receiveEdit.js
meissadia: From my testing we are only concerned about `1111`.  When entering `1111` in a Float field, it gets converted to `1111.0` which is covered by our current pattern.  With that in mind, `/^1111$/` seems too restrictive but `/^1111/` is probably still too loose of a match.  
 
 I think `/^1111(\.0)?/` covers the instances I've seen but I will double check with the backend folks to identify other edge cases. 👍🏽

lchen-2101 (approved):
LGTM

PR #1203, opened 2021-11-19
Wrap Alert child string in a Fragment

### LINKED ISSUE #1202: Crash when signing up for mailing list
Our common `Alert` component expects the item it wraps to be an element/component.

meissadia:
Closes #1202

I believe that because of the way our Alert component clones its children, we must wrap this `message` string in a Fragment.  Otherwise the Frontend crashes.

Screen Shot 2021-11-19 at 3 39 54 PM


lchen-2101 (approved):
LGTM

lchen-2101:
actually... can you make a note in the code of why we're doing the fragment?

lchen-2101 (approved):
LGTM

PR #1209, opened 2021-11-22
Q659-1: Highlight usage of 888 or 88888 in fields where 8888 == NA

### LINKED ISSUE #1199: Bold the impacted field in Q659
Testing Q659 on dev: It would be helpful to bold the field that is triggering the edit, so the user can immediately find the problem. 

image

Similar to the way we do for S303.

image

meissadia:
Closes #1199 
## Screenshots
2022_Q1~Q659-1

## Testing via Local Platform
- Note: This is a new edit for 2022
- Open filing on the Frontend for 2022-Q1 via `src/common/constants/dev-config.json`
  config
- Submit test file `cypress/fixtures/2022-Q1-FRONTENDTESTBANK9999-Q659-1.txt` for `2022-Q1`
- Observe `888`/`88888` highlighted in Q659-1 listing while `8888` left unhighlighted.


lchen-2101 (approved):
LGTM, I'm guessing one's coming for 659-2? that's the 999, and 99999 one

PR #1271, opened 2021-12-21
FFVT - Remove year checks

### LINKED ISSUE #1270: Remove FFVT year check
We've established that there is only one FFVT parser for all years, and we no longer accept user-input on which year the submitted file is for, so we no longer need to check for valid years in the FFVT parsing logic.

meissadia:
Closes #1270 

We've established that there is only one FFVT parser for all years, and we no longer accept user-input on which year the submitted file is for, so we no longer need to check for valid years in the FFVT parsing logic.


PatrickGoRaft (approved):
LGTM.

PR #1296, opened 2022-01-03
HMDA Help - Admin-selectable years

### LINKED ISSUE #1294: [Beta] HMDA Help - Auto allow access to manage Institutions for the upcoming year
Starting in Q3, allow HDMA Help users the option to create, edit, and delete Institutions for the upcoming year, to facilitate internal testing. 

**ex.** As of the start of `2021-Q3`, HH users would be able to create `2022` Institutions

meissadia:
Closes #1294

- More nuanced derivation of Admin-selectable years to automatically provide access 
- Adds unit tests for `getFilingYears`

This fix will resolve the current Cypress error on prod-beta in the Institutions test suite where we attempt to create an Institution for a year too far in the future.

### DIFF
--- a/src/common/constants/configHelpers.js
+++ b/src/common/constants/configHelpers.js
@@ -43,7 +43,7 @@ export const getFilingYears = (config, options = defaultOpts) => {
     const [year, quarter] = splitYearQuarter(config.defaultPeriod)
     if (year) {
       const upcomingYear = quarter !== 'Q3' ? year : parseInt(year, 10) + 1
-      years.add(upcomingYear)
+      years.add(upcomingYear.toString())
     }
   }
 
--- a/src/common/constants/configHelpers.js
+++ b/src/common/constants/configHelpers.js
@@ -33,15 +33,16 @@ export const getFilingYears = (config, options = defaultOpts) => {
     isVisible && years.add(splitYearQuarter(period)[0])
   })
 
-  // Additional inclusions for HMDA Help users to enable administrative tasks
-  // for future/closed filing periods.
+  // Additions for HMDA Help users
   if (withAdmin) {
     config.timedGuards.preview.forEach(adminPeriod =>
       years.add(splitYearQuarter(adminPeriod)[0])
     )
 
-    const nextYear = parseInt(splitYearQuarter(config.defaultPeriod)[0], 10) + 1
-    years.add(nextYear.toString())
+    // Starting in Q3, automatically enable institution management for the upcoming year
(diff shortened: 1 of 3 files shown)


src/common/constants/configHelpers.js
```diff
@@ -33,15 +33,16 @@ export const getFilingYears = (config, options = defaultOpts) => {
...
     config.timedGuards.preview.forEach(adminPeriod =>
       years.add(splitYearQuarter(adminPeriod)[0])
     )
 
-    const nextYear = parseInt(splitYearQuarter(config.defaultPeriod)[0], 10) + 1
-    years.add(nextYear.toString())
+    // Starting in Q3, automatically enable institution management for the upcoming year
+    const [year, quarter] = splitYearQuarter(config.defaultPeriod)
+    const upcomingYear = quarter !== 'Q3' ? year : parseInt(year, 10) + 1
+    upcomingYear && years.add(upcomingYear.toString())
```
lchen-2101, line 45: the `&&` trick should really be limited to jsx display templates; within logic flow `if` statements are still the standard; also for this instance, would there be a scenario year would be `null` or `undefined`? I think we can just directly do the `years.add`, right?

lchen-2101 (changes requested):

src/common/constants/configHelpers.js
meissadia: In the normal flow of our app, `upcomingYear` should not be `undefined`, though I did encounter this scenario while mocking out inputs for unit tests, so I thought it was a helpful guard to have. 
 
I've pushed an update to use `if`.

lchen-2101: gotcha, if that's the case, shouldn't we just wrap the 2 lines into a if year statement? e.g.:
```
if (year) {
  const upcomingYear = ...
  years.add(...)
}
```
since parseInt would also fail / NaN in that instance.

meissadia: Good call 👍🏽 
 
 Also, no need to convert `years` to strings and then back to ints and then back to strings, so I simplified to `years.add(upcomingYear)`

lchen-2101: actually... the `toString` is probably still needed, because `parseInt(...) + 1` would return an int, and we could get duplicate entries in the set for the same year in int form, and string form; the scenario probably wouldn't arise, but from purely programmatic perspective, it could be an issue.

meissadia: 👍🏽 `years.add(upcomingYear.toString())`

lchen-2101 (approved):
LGTM

PR #1299, opened 2022-01-04
Update privacy act link in mailing list signup to www for cf.gov

wpears:
The currently link is inaccessible off-network


meissadia (approved):
💥 thank you!

PR #1301, opened 2022-01-05
2022 test files - Update and census tract to resolve V627

### LINKED ISSUE #1300: 2022 test files - census tract
Update census tract to avoid V627

meissadia:
Closes #1300 

I thought we had fixed this in #1262 but this edit is popping up on Beta.  

Filing with both these files on Beta was able to reach the signing stage.


lchen-2101 (approved):
LGTM

PR #1313, opened 2022-01-20
Sets reasonable defaults for rate-spread

wpears:
## Changes

- Sets default loan term to 30 years
- Sets default rate set/lock in date to the current day

These are just my best guesses, I'd recommend a quick analysis of data submitted to `/public/rateSpread` to find out whether these defaults are the most sensible (I'm fairly confident on 30 years, not super confident on rate set date.

### DIFF
--- a/src/tools/rate-spread/Form.jsx
+++ b/src/tools/rate-spread/Form.jsx
@@ -13,7 +13,7 @@ const defaultState = {
   reverse: '2',
   rateSetDate: (new Date()).toLocaleDateString(),
   APR: '',
-  loanTerm: '30',
+  loanTerm: '',
   validationErrors: {},
   isFetching: false,
   error: false,


meissadia:
Thanks for working on this @wpears!

I'll take the analysis piece to the team and see what we can determine.

Some thoughts on the UI changes: 

Concerns
  - Could affect users who's scripts look for `mm/dd/yyyy` or `(1-50 years)`
    - Maybe this isn't a concern since integrations should be using the API?
      - "If you're using this page to automate, please consider using the Rate Spread API instead"
        - https://cfpb.github.io/hmda-platform/#rate-spread-single-ratespread
        - https://cfpb.github.io/hmda-platform/#rate-spread-batch-ratespreads
      - Experiment here with displaying the API call used by the tool/feature?

Alternatives
  - Buttons next to each field that offer users a sensible default or helpful adjustments?
    - Date: `back 1 week`, `forward 1 week`
    - Loan Term: `15 years`, `30 years`
    - APR%: none

src/tools/rate-spread/Form.jsx
```diff
@@ -11,9 +11,9 @@ const defaultState = {
   actionTaken: '1',
   amortization: 'Fixed',
   reverse: '2',
-  rateSetDate: '',
+  rateSetDate: (new Date()).toLocaleDateString(),
   APR: '',
-  loanTerm: '',
+  loanTerm: '30',
```
meissadia, line 16:
```suggestion
  loanTerm: '',
```

meissadia (changes requested):
Per [Eric's](https://github.com/cfpb/hmda-frontend/issues/1318#issuecomment-1021515666) feedback, let's default the `rateSetDate` but leave the loan term unset.

meissadia (approved):
LGTM 👍🏽 Thank you

PR #1329, opened 2022-01-27
[Filing] Block non-UTF8 encoded file uploads with message

### LINKED ISSUE #1311: Filing: Identify uploads with BOM, non-UTF8
Similar to the way we [check in the FFVT](https://github.com/cfpb/hmda-frontend/blob/4b5f1a5e89bb492b1ecb651f0777a1ab7ffc255b/src/tools/file-format-verification/containers/UploadForm.jsx#L33-L50), we can explore warning users of file encodings that tend to cause processing errors on the Platform.

meissadia:
Closes #1311

## Changes

- Checks file encoding of uploads `if (fileInfo.encoding !== 'UTF-8')`
- Moves other checks inside of `detectFileEncoding` promise for consistency. 

## Testing

1. I've run 5 1M row submissions through this check and have not found any memory spikes 

## Screenshots
Screen Shot 2022-01-27 at 12 42 23 PM

### DIFF
--- a/src/filing/utils/checkFileErrors.js
+++ b/src/filing/utils/checkFileErrors.js
@@ -48,7 +48,7 @@ function fileIsEmpty(file) {
 }
 
 function extensionIsNotTxt(file) {
-  let extension = file?.name?.split('.').slice(-1)[0]?.toLowerCase()
+  const extension = file?.name?.split('.').slice(-1)[0]?.toLowerCase()
   if (extension !== 'txt')
     return 'The file you uploaded is not a text file (.txt). Please check your file and re-upload.'
 }
--- a/src/filing/utils/checkFileErrors.js
+++ b/src/filing/utils/checkFileErrors.js
@@ -6,41 +6,54 @@ export default function checkFileErrors(file, cb) {
 
   reader.addEventListener('load', () => {
     const firstBytes = reader.result
-    const errors = []
-    
-    if (file && file.size !== undefined && file.name !== undefined) {
-      detectFileEncoding(file).then(fileInfo => {
-        if (fileInfo.encoding !== 'UTF-8'){
-          errors.push(
-            'The file you selected is not UTF-8 encoded. Please check your file and re-upload.'
-          )
-          
-          // Skip additional checks to avoid false-positives
-          return cb(errors)
-        }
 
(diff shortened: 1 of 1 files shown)


src/filing/utils/checkFileErrors.js
```diff
@@ -1,31 +1,46 @@
...
-        errors.push('Your file appears to be a text file (.txt) but has invalid content. Please ensure you are uploading a pipe-delimited text file and that your transmittal sheet begins with 1.');
-      }
+      detectFileEncoding(file).then(fileInfo => {
+        if (fileInfo.encoding !== 'UTF-8'){
+          errors.push(
+            'The file you selected is not UTF-8 encoded. Please check your file and re-upload.'
+          )
+          
+          // Skip additional checks to avoid false-positives
+          return cb(errors)
```
lchen-2101, line 19: return here feels awkward, knowing the purpose of it... feels like this section could be refactored a bit; reversing the `if` logic here might flow better; so something like:
```
if (fileInfo.encoding === 'UTF-8') {
    ... // the previous logic
} else {
    cb(...)
}
```
in some of `cb` calls that would only have 1 error, maybe just call `cb(['Your file was...'])` directly, and not have to do `errors.push`.

more of a personal preference addition refactoring, we might be able to do something like:
```
errors = fileSizeCheck(file, errors)
errors = txtCheck(file, errors)
...
```

meissadia: Agreed, that block is hard to decipher.  Refactored.  Thanks for always calling out hard-to-read code!

src/filing/utils/checkFileErrors.js
```diff
@@ -1,31 +1,59 @@
...
+    return 'The file you selected is not UTF-8 encoded. Please check your file and re-upload.'
+}
+
+function fileIsEmpty(file) {
+  if (file?.size === 0)
+    return 'The file you uploaded does not contain any data. Please check your file and re-upload.'
+}
+
+function extensionIsNotTxt(file) {
+  let extension = file?.name?.split('.').slice(-1)[0]?.toLowerCase()
```
lchen-2101, line 51: think `const` should work here, right?

meissadia: Update 👍🏽

lchen-2101 (approved):
LGTM

PR #1333, opened 2022-01-28
Fix cypress tests

meissadia:
## Changes

2022 & 2022-Q1
- Census tract
- Application date


meissadia:
Self-merging this PR so that I can roll these updates into the testing pod deployment tonight.

PR #1345, opened 2022-02-10
Open Frontend but closed Backend

### LINKED ISSUE #1344: Prevent fetchNewFiling for closed periods
Currently resulting in an error banner and endless loading spinner.
 
More info:

This occurs when an Institution attempts to access a closed period for which they do not already have a Filing.  We already utilize the frontend timed guards to skip creating new Filings for closed periods.  However, if the frontend is configured to have a Filing Period open, but the API is configured to have that period closed, the Frontend ends up stuck in an error state. 
 
This _should not_ happen and I'm struggling to figure out where I originally encountered this.  I think it was during a Maintenance window where we have a testing phase that opens the UI so that we can verify the system post-maintenance.

meissadia:
Closes #1344 

## Changes

If the Frontend attempts to create a new Filing for a period that is closed (i.e. When a filer has no Filing nor any submissions), and receives a status message that starts with `The provided year or quarter is no longer available`, we appropriately display that those institutions have no filings. 
 
This scenario should not occur as we already have Timed Guard based checks to avoid attempting to create Filings for closed periods, but this is a fail-safe for the case where the Frontend and Backend are out of sync.

## UI
Local Frontend pointed at Production
Screen Shot 2022-02-10 at 6 44 36 PM

### DIFF
--- a/src/filing/actions/fetchNewFiling.js
+++ b/src/filing/actions/fetchNewFiling.js
@@ -6,6 +6,7 @@ import { createFiling } from '../api/api.js'
 import { error } from '../utils/log.js'
 import receiveLatestSubmission from './receiveLatestSubmission'
 import receiveNonQFiling from './receiveNonQFiling'
+import { FilingNotAllowed } from '../../common/constants/platform-messages.js'
 
 export default function fetchNewFiling(filing) {
   return dispatch => {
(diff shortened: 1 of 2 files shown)


src/filing/actions/fetchNewFiling.js
```diff
@@ -13,6 +14,21 @@ export default function fetchNewFiling(filing) {
       .then(json => {
         return hasHttpError(json).then(hasError => {
           if (hasError) {
+            const ignoreMessage =
+              /^The provided year or quarter is no longer available|^Bad/
```
lchen-2101, line 18: might want to make this configurable; backend changes that shows this message might not be aware of it being used in the frontend; further down the line we'll need to think of a way to synchronize system messages.

lchen-2101 (approved):
LGTM

PR #1456, opened 2022-06-14
[Config] Never cache environment config

### LINKED ISSUE #1455: Avoid sticky config
To improve application configurability, without the need for full application redeployment, we use a Github hosted JSON configuration file.  
 
When we update this configuration, we've experience episodes where it takes 5+ plus minutes for the new configuration to be reflected in our Production app.  
 
I would like to explore adjusting the `cache` headers when `fetch`ing this config in an attempt to avoid the "sticky config" and by bypassing the caching of this file.

meissadia:
Closes #1455 
 
 ## Changes
Uses `fetch`'s `cache` option to always fetch the latest configuration from Github, bypassing the cache, in order to ensure that users receive the most up-to-date config in a timely manner. 
 
 https://developer.mozilla.org/en-US/docs/Web/API/Request/cache


meissadia:
Guess this wasn't needed.

PR #1471, opened 2022-06-23
Redirect from base urls

### LINKED ISSUE #1466: Redirect to latest year from base urls
- [x] Redirect to a year when none is provided in the URL 
    -  i.e.https://ffiec.cfpb.gov/data-browser/data/
    -  i.e.https://ffiec.cfpb.gov/data-publication/one-year-national-loan-level-dataset/

meissadia:
Closes #1466 

## Changes

When visiting Data Browser or Data Publications root urls, users will now be automatically redirected to the latest year

ex. https://ffiec.cfpb.gov/data-browser/data/ -> https://ffiec.cfpb.gov/data-browser/data/2021
 
ex https://ffiec.cfpb.gov/data-publication/one-year-national-loan-level-dataset/ -> https://ffiec.cfpb.gov/data-publication/one-year-national-loan-level-dataset/2021

### DIFF
--- a/src/common/redirectIfNoYearProvided.js
+++ b/src/common/redirectIfNoYearProvided.js
@@ -1,0 +1,21 @@
+import React from 'react'
+import { Redirect } from 'react-router-dom'
+
+/**
+ * Redirects users from the base URL, with no year provided, to a year focused URL
+ * @param {Object} render_props React Router props
+ * @param {String} targetYear Latest available year of Publication data
+ * @returns
+ */
+export const redirectIfNoYearProvided = (render_props, targetYear) => {
+  const year = render_props.match.params.year
+  let { pathname, search } = render_props.location
+
+  if (!year) {
+    if (pathname.substr(-1) !== '/') pathname = pathname + '/'
+
+    return <Redirect to={`${pathname}${targetYear}${search}`} />
+  }
+
+  return null
+}
--- a/src/data-browser/index.js
+++ b/src/data-browser/index.js
@@ -1,39 +1,30 @@
 import React from 'react'
-import { Switch, Route, Redirect } from 'react-router-dom'
+import { Switch, Route } from 'react-router-dom'
 import Home from './Home'
 import Geography from './datasets/Geography.jsx'
 import MapsGraphs from './maps/MapsGraphs.jsx'
 import NotFound from '../common/NotFound'
 import { withAppContext } from '../common/appContextHOC'
+import { redirectIfNoYearProvided } from '../common/redirectIfNoYearProvided'
 
 const DataBrowser = props => {
   const { publicationReleaseYear } = props.config
 
   return (
-    <div className="DataBrowser App">
+    <div className='DataBrowser App'>
       <Switch>
-        <Route exact path="/data-browser" component={Home} />
+        <Route exact path='/data-browser' component={Home} />
         <Route
-          path="/data-browser/data/:year?"
-          render={(r_props) => <Geography {...r_props} {...props} /> }
+          path='/data-browser/data/:year?'
+          render={r_props =>
+            redirectIfNoYearProvided(r_props, publicationReleaseYear) || <Geography {...r_props} {...props} />
+          }
         />
         <Route
           path={['/data-browser/maps/:year?', '/data-browser/maps']}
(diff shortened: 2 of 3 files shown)


src/data-browser/index.js
```diff
@@ -1,39 +1,30 @@
...
 import NotFound from '../common/NotFound'
 import { withAppContext } from '../common/appContextHOC'
+import { redirectIfNoYearProvided } from '../common/redirectIfNoYearProvided'
 
 const DataBrowser = props => {
   const { publicationReleaseYear } = props.config
 
   return (
-    <div className="DataBrowser App">
+    <div className='DataBrowser App'>
```
lchen-2101, line 14: why the quote style change? while this is jsx, think normally html style tag attributes should be double quotes?

src/common/redirectIfNoYearProvided.js
```diff
@@ -0,0 +1,21 @@
...
+  const year = render_props.match.params.year
+  let { pathname, search } = render_props.location
+
+  if (!year) {
+    if (pathname.substr(-1) !== '/') pathname = pathname + '/'
+
+    return <Redirect to={`${pathname}${targetYear}${search}`} />
+  }
+
+  return null
```
lchen-2101, line 20: not sure if null explicitly is necessary here; aren't null and undefined treated similarly in conditionals?

src/data-browser/index.js
```diff
@@ -1,39 +1,30 @@
...
       <Switch>
-        <Route exact path="/data-browser" component={Home} />
+        <Route exact path='/data-browser' component={Home} />
         <Route
-          path="/data-browser/data/:year?"
-          render={(r_props) => <Geography {...r_props} {...props} /> }
+          path='/data-browser/data/:year?'
+          render={r_props =>
+            redirectIfNoYearProvided(r_props, publicationReleaseYear) || <Geography {...r_props} {...props} />
```
lchen-2101, line 20: think maybe hoc would make code cleaner?

src/common/redirectIfNoYearProvided.js
meissadia: Removed

src/data-browser/index.js
meissadia: Refactored

src/data-browser/index.js
lchen-2101: sounds good; yeah, jsx should transpile it to correct quotes;
for reference, when you inspect an element in chrome, the double quotes isn't indicative of the source as the browser would convert it regardless:
image
becomes
image

src/common/withRedirectToTargetYear.js
```diff
@@ -0,0 +1,30 @@
...
+}
+
+
+/**
+ * HOC to follow a <Redirect />, if provided
+ * @param {Object} props 
+ * @returns
+ */
+export const withRedirect = WrappedComponent => props => {
+  if (props.redirect) return props.redirect
```
lchen-2101, line 27: I was thinking more having the redirect logic encapsulated in here, rather than still having to call it with the component; all the props needed for the logic should be available if I'm not mistaken?

meissadia: Refactored + added Cypress tests to validate

src/common/withYearValidation.js
```diff
@@ -0,0 +1,69 @@
+import React from 'react'
+import { Redirect } from 'react-router-dom'
+
+/**
+ * Reconstructs the URL to point to the targetYear
+ * @param {Object} props Includes props from React Router
+ * @param {String} targetYear Latest available year for Publication/product
+ * @returns
+ */
+export const formatPath = (props, targetYear) => {
```
lchen-2101, line 10: I wonder if this function is generic enough that it can be helpful elsewhere; if it is, I don't think this is the place for it since we probably won't remember to associate this function within this file name.
if the function is more specific to the hoc, I don't think we need to export it.

meissadia: Looking at the path formatting in other sections of the app, they each require their own specific formatting rules. So this will stay local. I've removed the `export`.

lchen-2101 (approved):
LGTM

Issue #1176, opened 2021-10-26
Generate 2022 Quarterly test submission files


meissadia:
- [x] Q1
- [x] Q2 -> Platform not recognizing Action Taken dates as being in Q2
- [x] Q3 -> Platform not recognizing Action Taken dates as being in Q3

PR #1483, opened 2022-07-01
[Cypress] Add 2022-Q2 test file

meissadia:
Part of #1176 
 
I was able to work through my issue regarding the building of the Cypress testing image, so that has been deployed.
 
In the meantime, I've manually tested this on Prod and Beta and things look good.


lchen-2101 (approved):
LGTM

PR #1484, opened 2022-07-07
[Docs] Panel - Add enumeration -1

### LINKED ISSUE #1482: Public Panel - Data Fields with Values and Definitions Missing (-1) documentation
The current documentation for the available codes for "Other Lender Code" field does not include -1 (negative one) in the write up. -1 is meant to signify a null/blank value and is a valid entry for this field. Would need to add that clarification to the documentation to avoid confusion with researchers scanning the panel data. See screenshot below 

image

meissadia:
Closes #1482 

Add the missing -1 enumeration for `other_lender_code` in the `Public Panel - Data Fields with Values and Definitions` documentation. 
 
## Screenshot
other-lender-code-neg-1


lchen-2101 (approved):
LGTM

PR #1550, opened 2022-09-08
[Links] Fixed supplemental guide links for 2021 and 2022.

Michaeldremy:
PR fixes links that point to wrong supplement guide documents.

2021 and 2022 supplemental guide links now link to the correct documents.


meissadia:
Looks good! Thanks for jumping on this one.

Issue #1531, opened 2022-08-30
[Graphs] Allow direct linking to a section


meissadia:
Currently users must start by viewing the `Graphs` section and then navigate to the `Filer Info` or `FAQ` sections. 
 
We want to add a url parameter that tracks the selected section of the page being viewed to allow users to directly navigate to any section.

PR #1553, opened 2022-09-09
[Graphs] Adds direct linking to filer info and FAQ tabs.

Michaeldremy:
Closes https://github.com/cfpb/hmda-frontend/issues/1531

Users will now be able to directly link to `/info/filers` and `/info/faq` tabs in graphs.


meissadia:
Closes #1531

src/data-browser/graphs/SectionSelector.jsx
```diff
@@ -1,26 +1,43 @@
...
+        title === "Graphs"
+          ? graphURL
+          : `/data-browser/graphs/quarterly/info/${title
+              .replace(" ", "-")
+              .toLowerCase()}`
+      }
     >
-      {title}
-    </button>
+      <button
```
meissadia, line 28: Does this still need to be a <button>?  Seems like it's interaction is no longer required since the <Link /> element now handles the redirection, instead of relying on the Component state.
 
 I think the <NavLink /> from `react-router` also accepts `className`, so the styling might also be easily transferable.

src/data-browser/graphs/quarterly/index.jsx
```diff
@@ -1,22 +1,32 @@
...
+import { SectionFAQ } from "./SectionFAQ.jsx"
+import { SectionFilerInfo } from "./SectionFilerInfo"
+import { SectionGraphs } from "./SectionGraphs"
+import { SectionSelector } from "../SectionSelector"
+import { Switch, Route } from "react-router-dom"
+import Error from "../../../common/Error"
+import "../graphs.css"
 
-const SectionOptions = ['Graphs', 'Filer Info', 'FAQ']
+const SectionOptions = ["Graphs", "Filer", "FAQ"]
```
meissadia, line 12: I still prefer `Filer Info` over `Filer` for the nav-link label.  I think it's more descriptive of what the user will find in that section, even it doesn't exactly match the URL endpoint.

src/data-browser/graphs/quarterly/SectionGraphs.jsx
```diff
@@ -76,57 +92,104 @@ export const SectionGraphs = ({
...
+      )
+    }
+
+    // Allows direct linking to /info/filer or /info/faq and doesn't trigger a url update
+    // Additonally it stores the graph url and is used when user clicks the Graphs tab
+    if (props.history.location.pathname.includes("/info/filer")) {
+      props.history.push({
+        pathname: `${BaseURLQuarterly}/info/filer`,
+      })
+      dispatchGraphURL(
```
meissadia, line 132: `dispatchGraphURL` call is repeated for every condition which means it's not conditional.  Maybe have one call after the `if-else` block since it is currently called *after* the `history.push` in every case?

src/data-browser/graphs/quarterly/index.jsx
```diff
@@ -37,8 +47,18 @@ export const QuarterlyGraphs = props => {
             history: props.history,
           }}
         />
-        <SectionFilerInfo show={section === 'Filer Info'} />
-        <SectionFAQ show={section === 'FAQ'} />
+
+        <Switch>
+          {/* Setting direct paths to access other tabs */}
+          <Route
+            path={"/data-browser/graphs/quarterly/info/filer"}
```
meissadia, line 54: How would you feel about the route being `/info/filers`, since the page presents info about multiple filers?
 
A section label of `Filer Info` still makes sense to me, I think it just sounds better than `Filers Info`.  Maybe the linguistically correct version would be `Filers' Info`?

meissadia (changes requested):
Can you also work to recapture the pill-like styling for the section selector?
|current|prod|
|--|--|
|Screen Shot 2022-09-14 at 2 46 11 PM|Screen Shot 2022-09-14 at 2 46 18 PM|

src/data-browser/graphs/quarterly/SectionGraphs.jsx
Michaeldremy: Good catch, I can't remember why I added the dispatch method to every `if-else` block. I updated the changed to dispatch outside all the `if-else`.

src/data-browser/graphs/quarterly/index.jsx
Michaeldremy: The `SectionOptions` array controls the label name and what appears in the URL. 

If we revert back to `Filer Info` the URL will say `info/filer-info`. Are you okay with it saying this?

src/data-browser/graphs/quarterly/index.jsx
Michaeldremy: Any changes to this would change the tab label and update the URL. How would `Filers' Info` appear in the year? Look above for previous comment about this.

src/data-browser/graphs/SectionSelector.jsx
Michaeldremy: The `<Link />` component from `react-router` is responsible for handling the URL being updated with the respective tab.

If I remove the `<Link />` the URL no longer gets updated upon a tab selection. 

I tried removed `className` from the `<button />` to the `<Link />` component and it yielded interesting results.

image

meissadia: Sorry, I didn't realize my element tags had been removed from the comment.
 
 I was asking if the `children` of the \<Link /> component needed to be a \<button>, since the \<Link /> is now handling the navigation. 
 
Also it looks like react-router's \<NavLink /> might be able to better apply the className?  Also review the styles in [SectionSelector.css](https://github.com/cfpb/hmda-frontend/blob/master/src/data-browser/graphs/SectionSelector.css) to see if you can adjust them to apply to the new anchor elements.

src/data-browser/graphs/quarterly/index.jsx
meissadia: Could you extend the logic to only use the first word of the label as the URL reference? (split at space, take first word, remove any non-alphabetic characters).

Or create a map from `label => url`?

src/data-browser/graphs/quarterly/index.jsx
meissadia: As above, we can work to change the derivation of the URL from the label to be a mapped relationship. 
```
[{label: 'Graphs', url: '/graphs'}, {label: 'Filers Info', url: '/filers'}, {label: 'FAQ', url: '/faq'}]
```
 
 Then adjust the intermediary logic as needed.

Michaeldremy: Updated code to use array of objects rather then array of strings to allow for more dynamic content control.

src/data-browser/graphs/quarterly/index.jsx
Michaeldremy: This has been fixed by using an object instead of a string.

`{label: 'Filers Info', url: '/filers'}`.

src/data-browser/graphs/SectionSelector.jsx
Michaeldremy: Resolved.

src/data-browser/graphs/quarterly/index.jsx
```diff
@@ -1,33 +1,45 @@
...
       <HomeLink />
       <GraphsHeader overview={graphHeaderOverview} />
       <Error error={error} />
-      <SectionSelector
-        options={SectionOptions}
-        selected={section}
-        onChange={setSection}
-      />
-      <div className='section-wrapper'>
+      <SectionSelector options={SectionOptions} selected={section} />
```
meissadia, line 39: `<SectionSelector>` should be able to internally determine which section is selected.

src/data-browser/graphs/quarterly/index.jsx
```diff
@@ -37,8 +49,20 @@ export const QuarterlyGraphs = props => {
             history: props.history,
           }}
         />
```
meissadia, line 51: This can be moved into the `<Switch>` below as kind of a default route.  It would be the last `<Route>` in the `<Switch>` and it wouldn't need a `path`.  
 
 That would also allow us to remove `show={section === 'Graphs'}`. 
  
  ```
            <Route
            render={() => (
              <SectionGraphs
                {...{
                  error,
                  setError,
                  setGraphHeaderOverview,
                  location: props.location,
                  match: props.match,
                  history: props.history,
                }}
              />
            )}
          />
        </Switch>
        ```

src/data-browser/graphs/quarterly/index.jsx
```diff
@@ -37,8 +49,20 @@ export const QuarterlyGraphs = props => {
...
+          {/* Setting direct paths to access other tabs */}
+          <Route
+            path={"/data-browser/graphs/quarterly/info/filers"}
+            render={() => (
+              <SectionFilerInfo show={section.label === "Filers' Info"} />
+            )}
+          />
+          <Route
+            path={"/data-browser/graphs/quarterly/info/faq"}
+            render={() => <SectionFAQ show={section.label === "FAQ"} />}
```
meissadia, line 63: With the above changes, we can remove the `show` prop since the `path` now determines which `<Section>` should be displayed.  
 
Inside of each `<Section>` we can also remove the conditional rendering using `if(!show)`.  And since that functionality was the main reason for splitting their code into Components, we can probably flatten the hierarchy and replace `SectionFilerInfo/SectionFAQ` with the content that they render.

src/data-browser/graphs/quarterly/index.jsx
```diff
@@ -1,33 +1,45 @@
...
+import { SectionGraphs } from "./SectionGraphs"
+import { SectionSelector } from "../SectionSelector"
+import { Switch, Route } from "react-router-dom"
+import Error from "../../../common/Error"
+import "../graphs.css"
 
-const SectionOptions = ['Graphs', 'Filer Info', 'FAQ']
+const SectionOptions = [
+  { label: "Graphs" }, // url for graphs comes from redux
+  { label: "Filers' Info", url: "/filers" },
```
meissadia, line 14: Sorry, I keep going back and forth on this, but maybe we should leave the label as `Filer Info` 😬 .  Even if `Filers' Info` is more accurate, it seems like `Filer Info` is just easier to read and reason about. Thoughts?

meissadia (changes requested):

src/data-browser/graphs/quarterly/index.jsx
meissadia: Sorry for sending you down that rabbit hole.  I also tried to debug this but couldn't figure out exactly which piece was doing the resetting.  I've refactored to eliminate the unmounting/mounting when switching from/to the Graph tab.

src/data-browser/graphs/quarterly/index.jsx
meissadia: I flattened things out and deleted the SectionFilerInfo/SectionFAQ components.

meissadia (approved):
@Michaeldremy Looks good! Please do one last round of testing to verify that the update I pushed for the `visibleSeries` reset does what it's supposed to. 🙏🏽

PR #1584, opened 2022-10-04
[Homepage] Make banner a consistent height to avoid controls shifting…

### LINKED ISSUE #1583: [Homepage] Stabilize banner height
- [ ] Set banner to be a consistent height to avoid shifting of controls

meissadia:
Closes #1583 

Issues
- [x] Scroll bars always show even without overflow
  - Chrome 105.0.5195.102
  - Chrome 105.0.5195.125 (looks good)
  - RESOLUTION: Removed the `overflow: scroll`.  
    - This does mean that we're restricted to 2 rows of text per banner message.

### DIFF
--- a/src/common/OptionCarousel.css
+++ b/src/common/OptionCarousel.css
@@ -7,6 +7,15 @@
   align-items: center;
 }
 
+/* Prevent shifting height on homepage banner  */
+#oc-homepage-banner .alert-text {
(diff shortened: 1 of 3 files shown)


Michaeldremy (approved):
Pulled down the branch and looked locally. Looks good to me!

PR #1602, opened 2022-10-07
Rtk query update

### LINKED ISSUE #1528: [Graphs] RTK Query for simplified data fetching
Utilize RTKQ to reduce the amount of boilerplate code needed to handle data pulled in from the Graphs API and stored in Redux.
 
https://redux-toolkit.js.org/tutorials/rtk-query

lchen-2101:
closes #1528

### DIFF
--- a/src/data-browser/graphs/SectionSelector.jsx
+++ b/src/data-browser/graphs/SectionSelector.jsx
@@ -8,8 +8,8 @@ import { useEffect, useState } from "react"
 const SectionOption = ({ isSelected, title, url }) => {
   const sectionClasses = `button section ${isSelected && "selected"}`
 
-  const graphStore = useSelector(({ graphs }) => graphs)
-  const graphURL = graphs.getConfig(graphStore, GRAPH_URL) // Getting graph url string from redux store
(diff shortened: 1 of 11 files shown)


meissadia (approved):
Awesome! Really simplifies things, thanks for this.  
 
 I'll wait until @Michaeldremy has a chance to review and ask questions before we merge.

Michaeldremy (changes requested):
Code looks great! I agree with Meissa it really simplifies everything. I did notice a visual bug that I am seeing. See my thought below.

I pulled down the branch locally and was doing manual testing and I found a potential issue.

Observations:

While viewing the graphs on the production website and cycling through the graphs, Highcharts will smoothly transition from one graph to another.

Issue:

When you select a different graph besides the initial one that is pre-loaded it starts to fill with the old graph information for a brief second then it loads the graph you selected. This is causing a **jolt** like effect.

After viewing multiple graphs and then traversing back it transitions smoothly again (since we are caching the request/data of the already viewed graphs)

lchen-2101:
> Member

We've jumped on a call and saw the both branches behave the same way while running locally.

Michaeldremy (approved):
LGTM

PR #1609, opened 2022-10-12
[Graphs] Y-Axis Labels displays M or K on respective values

### LINKED ISSUE #1560: [Graphs] Display Y-axis labels in the appropriate units
Labels for graphs with y-values in the millions are being displayed as ex. 1000K instead of 1M. 
 
Explore building a custom formatter function to remedy this. 
https://www.highcharts.com/forum/viewtopic.php?t=34647

Michaeldremy:
Closes #1560

### DIFF
--- a/src/data-browser/graphs/highchartsConfig.js
+++ b/src/data-browser/graphs/highchartsConfig.js
@@ -1,17 +1,17 @@
-import { hmda_charts, seriesColors, yearQuarters } from './config'
-import { cloneObject, filterByPeriods } from './utils/utils'
+import { hmda_charts, seriesColors, yearQuarters } from "./config"
+import { cloneObject, filterByPeriods } from "./utils/utils"
 
 // Highcharts configuration for a line graph
 export const baseConfig = {
   title: {
-    text: 'Title',
+    text: "Title",
   },
   subtitle: {
-    text: 'Subtitle',
-    style: { fontSize: '15px' },
+    text: "Subtitle",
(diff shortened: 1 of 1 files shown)


src/data-browser/graphs/highchartsConfig.js
```diff
@@ -151,25 +152,39 @@ export const deriveHighchartsConfig = ({
...
+      formatter: (tick) => {
+        if (!tick) return
+
+        if (tick.value > 999 && tick.value < 1000000) {
+          return (tick.value / 1000).toFixed(0) + "K" // convert to K for number from > 1000 < 1 million
+        } else if (tick.value >= 1000000) {
+          return (tick.value / 1000000).toFixed(1) + "M" // convert to M for number from > 1 million
+        } else if (tick.value < 900) {
+          return tick.value // if value < 1000, nothing to do
+        }
```
meissadia, line 167: What happens if the value is `950`?
 
I don't think we need any comparison here.  Just, in all other cases, return the plain value. 
```suggestion
        } else {
          return tick.value // if value < 1000, nothing to do
        }
```

meissadia (changes requested):
Looks pretty good.  One change.

src/data-browser/graphs/highchartsConfig.js
Michaeldremy: Code has been updated!

meissadia (approved):
Looks good!

PR #1612, opened 2022-10-13
Add configuration to use proof key for code exchange

kgudel:


meissadia (changes requested):
There's also a `keycloak.init` in [HMDA Help](https://github.com/cfpb/hmda-frontend/blob/master/src/hmda-help/index.js#L50)

PR #1642, opened 2022-11-03
1628 refactor cypress image

### LINKED ISSUE #1627: [Cypress] Decouple list of tests from Docker image
- Decouple which tests are run from the Cypress pod so that we don't need to rebuild/deploy the multi-GB image 
  - Similar to prod-config.json / prod-beta-config.json but for Cypress?
  - Fetch config from Github?  
  - Fetch load-test file from S3?
  - Provide the list of tests as a environment variable to the Pod  🤔 
    - CYPRESS_INTEGRATION_TEST_LIST: regex | string
    - CYPRESS_LOAD_TEST: true | false

### LINKED ISSUE #1628: [Cypress] Re-evaluate image build process
## Issue
In order to provide our team better control over CVEs Barak built our Cypress testing image from a raw CentOS layer.  
 
The downside is that each new build results in 5GB+ of new Docker layers. 
 
My hope is that we can find a way to streamline this build in order to:
1. reduce build time
2. drastically reduce upload time and bandwidth

## Thoughts
- Can we utilize a pre-built? 
  - https://docs.cypress.io/examples/examples/docker#Images
  - Have they improved their CVE upgrade cycles?
- Can we pull the `docker-runner.sh` script into the Pod from Github at runtime?
  - Sounds like something that would get flagged as a security vulnerability
  - Would allow us to use the same Cypress pod but update it's contents (fixtures, spec files, which tests run) dynamically
- Build our own CentOS + Cypress base image
  - Use ^^ as the basis for building the final image, which would include Fixtures, run script, etc.
  - Again, could we dynamically pull fixtures, list of tests into the pod at runtime or is this too risky?
- Re: Dynamic pulling of content
  - We already do this for the LargeFiler LAR file, so it seems like this approach is reasonable
  - Pull `docker-runner.sh` from Github:master?
  - Pull fixtures from Github:master 
  - Would pulling from S3 be more secure?

meissadia:
Closes #1628 
Closes #1627 

Changes to support new process of hosting specs/fixtures on S3 and pulling those files into the Cypress test pod at runtime in order to enable dynamic updates to test data without image rebuild.

Also adds new ENV variables to control which tests are run. 
 
New process:
1) Run cypress/update-s3-files.sh to upload test data and specs from the current local branch to S3.
2) Define ENV vars as outlined in cypress/README.md to configure which tests will run
3) Build and push Cypress test image (only needed for Cypress version updates or security fixes)

### DIFF
--- a/cypress/README.md
+++ b/cypress/README.md
@@ -45,6 +45,8 @@ export CYPRESS_PASSWORD=<password>
 export CYPRESS_INSTITUTION=<target-institution-lei>
 export CYPRESS_TEST_DELAY=1000 # No delay (milliseconds)
 export CYPRESS_ACTION_DELAY=1000 # No delay (milliseconds)
+export CYPRESS_SHOULD_LOAD_TEST=true
+export CYPRESS_TEST_LIST='cypress/integration/data-browser/**,cypress/integration/data-publication/**,cypress/integration/filing/**,cypress/integration/hmda-help/**,cypress/integration/tools/**'
 ```
 
 ## Running Tests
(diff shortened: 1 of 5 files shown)


Michaeldremy (approved):
LGTM

PR #1643, opened 2022-11-04
[Filing] Updated submission receipt to reflect eastern timezone

### LINKED ISSUE #1409: Add timezone to Submission receipt
- Maybe this becomes convert times/dates to ET?
- Since the Platform operates on ET, a West Coast user could have a Submission receipt that appears to be from a time/date before a Filing period is open.

Michaeldremy:
Closes #1409

### DIFF
--- a/src/filing/submission/Receipt.jsx
+++ b/src/filing/submission/Receipt.jsx
@@ -5,27 +5,38 @@ import PrintPageButton from './PrintPageButton'
 import { isBeta } from '../../common/Beta.jsx'
 import { ordinalHour } from '../utils/date.js'
 import { SIGNED } from '../constants/statusCodes.js'
+import { parseTimedGuardDate } from '../../deriveConfig.js'
 
-const Receipt = ({
-  status,
-  timestamp,
-  receipt,
-  filingPeriod,
-  email
-}) => {
+const Receipt = ({ status, timestamp, receipt, filingPeriod, email }) => {
+  const options = {
+    year: 'numeric',
+    month: 'long',
+    day: 'numeric',
+    timezone: 'America/New_York',
+  }
+
+  let daySubmitted = new Date(timestamp).toLocaleDateString('en-US', options)
+
+  let easternTime = new Date(timestamp).toLocaleString('en-US', {
+    timeZone: 'America/New_York',
+  })
+
+  let timeOfReceipt = daySubmitted + ',' + easternTime.split(',')[1] + " ET"
+
   const code = status.code
   if (code !== SIGNED || receipt === null) return null
   if (isBeta()) {
     return (
-      <section className="RefileWarning">
-        <Alert type="warning" heading="Beta filing complete">
+      <section className='RefileWarning'>
+        <Alert type='warning' heading='Beta filing complete'>
           <div>
-            You have successfully completed your <strong>beta</strong> HMDA filing for{' '}
-            {filingPeriod}.
+            You have successfully completed your <strong>beta</strong> HMDA
+            filing for {filingPeriod}.
             <br />
-            <strong>You will still need to file your final HMDA data to the{' '}
-               <a href="https://ffiec.cfpb.gov/filing">live system</a>{' '}
-               when the filing period is open.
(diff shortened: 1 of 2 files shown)


src/filing/submission/Receipt.jsx
```diff
@@ -5,47 +5,58 @@ import PrintPageButton from './PrintPageButton'
...
+    timezone: 'America/New_York',
+  }
+
+  let daySubmitted = new Date(timestamp).toLocaleDateString('en-US', options)
+
+  let easternTime = new Date(timestamp).toLocaleString('en-US', {
+    timeZone: 'America/New_York',
+  })
+
+  let timeOfReceipt = daySubmitted + ',' + easternTime.split(',')[1] + " ET"
```
meissadia, line 24: Let's extract this into a `formatReceiptTime` function. Then on line 54 we can just 
```
<strong>{formatReceiptTime(timestamp)}</strong>
```
 
 I would parse the date once, then reuse: `const date = new Date(timestamp)`

meissadia (changes requested):

src/filing/submission/Receipt.jsx
```diff
@@ -5,47 +5,58 @@ import PrintPageButton from './PrintPageButton'
 import { isBeta } from '../../common/Beta.jsx'
 import { ordinalHour } from '../utils/date.js'
```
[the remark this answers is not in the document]
Michaeldremy: I removed `import { ordinalHour } from '../utils/date.js'` but left in `import { isBeta } from '../../common/Beta.jsx'` due to 

```JavaScript
if (isBeta()) on line 12
```

src/filing/submission/Receipt.jsx
```diff
@@ -5,47 +5,58 @@ import PrintPageButton from './PrintPageButton'
...
 import { SIGNED } from '../constants/statusCodes.js'
+import { parseTimedGuardDate } from '../../deriveConfig.js'
```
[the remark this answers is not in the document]
Michaeldremy: Has been removed.

src/filing/submission/Receipt.jsx
Michaeldremy: Logic has been extracted from `Receipt` component and now resides in `utils/date` as a function.

src/filing/utils/date.js
```diff
@@ -68,4 +69,29 @@ export const isDstObserved = (date = new Date()) => {
...
+export const formatReceiptTime = timeToFormat => {
+  let timezone = {
+    timeZone: 'America/New_York',
+  }
+
+  const options = {
+    year: 'numeric',
+    month: 'long',
+    day: 'numeric',
+    timezone,
```
meissadia, line 88: This probably needs to be spread.  As you currently have it, you would end up with a nested object at the key `timezone`.
```suggestion
    ...timezone,
```

src/filing/utils/date.js
```diff
@@ -68,4 +69,29 @@ export const isDstObserved = (date = new Date()) => {
...
+  }
+
+  return (
+    new Date(timeToFormat).toLocaleDateString('en-US', options) +
+    ',' +
+    new Date(timeToFormat).toLocaleString('en-US', timezone).split(',')[1] +
+    ' ET'
+  )
```
meissadia, line 96:
```suggestion
  const date = new Date(timeToFormat)
  return (
    date.toLocaleDateString('en-US', options) +
    ',' +
    date.toLocaleString('en-US', timezone).split(',')[1] +
    ' ET'
  )
```

meissadia (changes requested):

src/filing/utils/date.js
Michaeldremy: Has been updated!

src/filing/utils/date.js
Michaeldremy: You are correct. Code has been updated.

meissadia (approved):
Verified on the East Coast by Abhi 🙏🏽 
 
Good to go!

PR #1694, opened 2022-12-22
[Cypress] Remove logout calls

### LINKED ISSUE #1693: [Cypress] Remove logout() calls
With the implementation of "test isolation" in Cypress, we no longer need to logout in-between authenticated tests.
 
Additionally, calls to `cy.logout` will cause test failures when there is no authenticated user.  
 
The error message indicates we can pass `failOnStatusCode: false` to avoid this behavior, however, `cypress-keycloak` (which implements the [cy.logout](https://github.com/babangsund/cypress-keycloak/blob/master/src/logout.ts) method) does not support the addition of this parameter.

meissadia:
Closes #1693 

Logout calls are not interfering with testing via our nightly cronjob but are a blocker when running from localhost.  Cypress has implemented `test isolation` which makes these logout calls redundant.  

## Changes
- Remove calls to cy.logout
- Remove unused param in cy.hmdaLogin
- Remove unused environment variable

### DIFF
--- a/cypress/e2e/filing/Filing.spec.js
+++ b/cypress/e2e/filing/Filing.spec.js
@@ -21,9 +21,6 @@ const years = (YEARS && YEARS.toString().split(',')) || getFilingPeriods(config)
 const { filingPeriodStatus } = config
 
 describe("Filing", function() {
-  // Only need to provide an Auth URL when running locally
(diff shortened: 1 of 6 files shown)


ojbravo (approved):
Looks good.

PR #1715, opened 2023-01-26
Graphs: Derive year for Filers from graph data

### LINKED ISSUE #1711: [Graphs] Filer Info table incorrectly displaying data
`Filers Info` tab found in [Graphs](https://ffiec.cfpb.gov/data-browser/graphs/quarterly/info/filers) is showing up blank data for 2023. 

### Issue: 

Table is attempting to display 2023 data but that data is not in the API response and can't be calculated since the 2023 filer data has not been finalized.

image

meissadia:
Closes #1711 

## Changes
Derive year for Filers from graph data.

If the Graphs API says 2023-Q1 data is the latest available, then filer info for 2023 should also be ready. 
When latest is 2022-Q3 or 2022 annual, we use 2022.


lchen-2101 (approved):
LGTM

PR #1763, opened 2023-04-11
[Synk] Upgrades Dependencies from Synk

Michaeldremy:
Closes #1741 

Packages that Synk requested to be upgraded, has been upgraded, audited for vulnerabilities and had some cypress tests updated due to a few tests failing.


ojbravo (approved):
Looks good.

PR #1770, opened 2023-04-14
3688 rate spread announcement

Michaeldremy:
PR is designed to display rate spread calculator announcement banner updates to prod and dev and add a new note in the update notes section.


ojbravo (approved):
Looks good to me!

PR #1785, opened 2023-05-12
[olarft] Fix harded olarft values

Michaeldremy:
Found a few files that had hard coded values for the olarft branch. This PR fixes those issues.


meissadia (approved):
👍🏾

PR #1787, opened 2023-05-16
Uswds-navbar

ojbravo:
## Changes

Adding USWDS + Algolia Search into FE

## Testing

Tested on hmda4


ojbravo:
Deleting this PR. Will be replaces with v2.

PR #1801, opened 2023-06-01
Fixes a bug inside DocLink component.

Michaeldremy:
PR fixes a bug that was found inside the DocLink component causing the Data browser table to not load.


ojbravo (approved):
Looks good.

PR #1815, opened 2023-06-12
Opens up testing for Aggregate & Disclosure reports

Michaeldremy:
Allows dev environment to handle 2022 testing for Aggregate and & Disclosure reports.


ojbravo (approved):
Simple update. Looks good!

PR #1822, opened 2023-06-23
removing banner for db updates

PatrickGoRaft:
## Changes

- Updates to constants JSON that remove maintenance banner for the Data Browder UI


ojbravo (approved):
Perfect.

PR #1845, opened 2023-07-11
HMDA FIG Prototype Banner

Michaeldremy:
Generates a banner that tells folks about the HMDA Online FIG prototype and adds a new link on beta site to allow people to directly go to the the HMDA FIG.


ojbravo (approved):
LGTM

PR #1846, opened 2023-07-12
Updated Readme: Check Digit

ojbravo:
Added terminal command to run Check Digit locally so you can then generate the MAX.txt file.


Michaeldremy (approved):
LGTM

PR #1849, opened 2023-07-14
[Filer Info] Heading content change

Michaeldremy:
Removes the yearly specific year from the header on the [filer info tab](https://ffiec.cfpb.gov/data-browser/graphs/quarterly/info/filers) on graphs.


ojbravo (approved):
Perfect.

PR #1938, opened 2023-09-28
[Filing] Introduces profile page

### LINKED ISSUE #1930: Complete Profile Page
- Create account page that enables users to update their user details including: first name, last name, and associated institutions
- The associated institutions selection should use the same logic as the registration theme in keycloak
- Update filing page to include a link to the new account page
- Create new frontend logic that automatically sends users to the new account page if their account does not have any associated institutions
- Write new Cypress tests

Michaeldremy:
Closes #1930


ojbravo (approved):
Amazing Work!

PR #1959, opened 2023-10-12
Opens future filing periods on DEV

Michaeldremy:
Opens 2023, 2024, 2024-Q1 filing periods for DEV. Also included in this PR is updating the `start` and `build` scripts in `package.json` which adds a new flag that is required when running node v17+


ojbravo (approved):
Looks good.

PR #2020, opened 2023-12-01
Announces MFA

Michaeldremy:
Announces MFA/Profile page and updates change-log to include the announcement.


ojbravo (approved):
Good work.

PR #2088, opened 2024-02-09
Removed 'trusted types' from nginx CSP

ojbravo:
Removed ```require-trusted-types-for 'script'";``` from the CSP in the nginx.conf as it was blocking the entire site from rendering.


Michaeldremy (approved):
LGTM

PR #2124, opened 2024-03-19
Reorders uswds loading scripts and reduced timeout

ojbravo:
USWDS navbar is now available after just 1 second.


Michaeldremy (approved):
LGTM

PR #2149, opened 2024-04-05
[Filing Profile Page] Adjust profile page due to Keycloak upgrade

### LINKED ISSUE #2147: [Filing/Profile] Fix Alert that tells users they need an institution to use filing
New version of Keycloak can now _not_ send the `LEI` key. Previous versions always sent the `LEI` key and based off this key from Keycloak we would display an alert to the user that you need to select one institution in order to use the filing platform. The alert depended on the `LEI` key.

Michaeldremy:
Closes #2147


ojbravo (approved):
Looks good!

PR #2154, opened 2024-04-10
[Cypress] Update Keycloak test

### LINKED ISSUE #2152: [Cypress] Fix failing tests
There are a few Cypress tests that are failing. Created a ticket to track the failing tests.

- [x] Keycloak: Can log in and out (Failing due to URL containing more information due to Keycloak upgrade)~
- [x] HMDA Help Institution (Can delete and create Institutions)~~

~ Keycloak was recently updated and with that update it has started to append `#iss=https%3A%2F%2Fffiec.cfpb.gov%2Fauth%2Frealms%2Fhmda2` to the URL. The test is using Regex and due to that Regex it fails because of the extra stuff added to the URL - **Test has been fixed by no longer doing a hard match on the URL but rather ensuring the year from the URL is included in our opened filing periods**

~~ When running test against production the test passes but the Cypress test runner is showing that the test is failing with the following error `expected '<td>' to contain '2023'` . The same error is successful when running again production - **Test passed last night following an update to the Cypress Docker Image**

Michaeldremy:
Closes #2152

### DIFF
--- a/cypress/e2e/filing/Keycloak.spec.js
+++ b/cypress/e2e/filing/Keycloak.spec.js
@@ -1,6 +1,8 @@
+import { getDefaultConfig } from '../../../src/common/configUtils'
+import { getFilingPeriods } from '../../../src/common/constants/configHelpers'
 import { isCI } from '../../support/helpers'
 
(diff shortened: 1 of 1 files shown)


ojbravo (approved):
Looks good.

PR #2156, opened 2024-04-11
Bug fix for Graphs dateRange selector

ojbravo:
Updated dateRange logic to limit periodLow options to less than periodHigh selection. Closes 2153.


Michaeldremy (approved):
LGTM!

Issue #2192, opened 2024-05-14
CSP Blocking dap.digitalgov.gov


ojbravo:
The following error is showing in the browser console: ```Refused to load the script 'https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=CFPB' because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-inline' blob: data: https://tagmanager.google.com https://www.googletagmanager.com https://www.google-analytics.com https://*.cfpb.gov https://www.consumerfinance.gov https://cdn.mouseflow.com". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.```

PR #2194, opened 2024-05-16
Removed unsafe-inline from CSP. Whitelisted dap Analytics.

ojbravo:
Closes #4318 in hmda-devops and #2192 in hmda-frontend


Michaeldremy (approved):
LGTM

PR #2249, opened 2024-07-31
[Profile Page] Auth Token

### LINKED ISSUE #2121: Create Widget for Getting Token from Keycloak using Login.gov
As we move into using keycloak + login.gov exclusively for authentication we need an easy way for internal and external users to get an authentication token. It seems like the easiest intitial way of accomplishing this is to create a stripped down app that routes users to keycloak and then outputs the token when keycloak routes the user back.

Michaeldremy:
Closes #2121 

Adds the ability for any vendor to get their auth token.

image


ojbravo (approved):
Nicely done.

PR #2258, opened 2024-08-12
[Graphs] Data Table Expandable & Data Issue Note

### LINKED ISSUE #2252: Update Graphs to Include explainer about small sample size removal.
Add the following explainer text below the graphs:

`Note: Data points which would be generated based on fewer than 100 loans are not shown. This may cause gaps in graphs.`

Michaeldremy:
Closes #2252 

The automatic `HighCharts` data table is now being handled manually. The reason for this is to allow the data table styling to be better controlled. 

The data table under the graph is now expandable (closed by default).

image

Additionally a note about the gaps in the graph and table now appears in the UI under the custom data table.

image

Tests are passing against DEV ✅

### DIFF
--- a/src/common/ExpandableSection.css
+++ b/src/common/ExpandableSection.css
@@ -39,3 +39,8 @@
   padding: 1em 1.3em;
   width: 100%;
 }
+
+/* Table CSS within Graphs */
+.expandable-section .bold-font {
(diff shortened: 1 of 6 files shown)


ojbravo (approved):
Works great on dev!

PR #2288, opened 2024-09-24
[BETA] Temporary Fix for `/filing` navigation

Michaeldremy:
Updated `<Header />` component logic to use the `isBeta` flag to use an `a` tag HTML element or `Link` from `react-router-dom` to resolve an issue where navigating from the homepage on BETA to `/filing` was causing it to crash.


ojbravo (approved):
Looks good.

PR #2384, opened 2025-01-06
[Config] update dev-beta

Michaeldremy:
The `dev-beta` config has not been updated in a few years. The following PR will make the `dev-beta` config match similarly to our dev config.


ojbravo (approved):
Looks good.

PR #2385, opened 2025-01-07
Proof of concept showing the frontend using keycloak on an arbitrary domain

kgudel:


src/common/api/Keycloak.js
```diff
@@ -28,12 +38,12 @@ export const initKeycloak = (overrides) => {
     } else if (import.meta.env.MODE === 'development') {
       keycloak = new Keycloak('/local_keycloak.json')
     } else {
-      keycloak = new Keycloak('/keycloak.json')
+      keycloak = new Keycloak(keycloakConfig)
```
tptignor, line 41: Would it be any easier to manage/revert the keycloak config as a ConfigMap?

src/common/api/Keycloak.js
```diff
@@ -4,6 +4,16 @@ let keycloak = null
 let isInitialized = false
 let initPromise = null
 
+const keycloakConfig = {
+  "realm": "hmda2",
+  "url": "https://{{domain}}}/auth",
```
ojbravo, line 9: Can you try replacing line 9 with the following to make it dynamic?
```"url": `https://${window.location.hostname}/auth`,```

ojbravo: To account for local testing, please use the following:
```const hostname = window.location.hostname
let keycloakRedirect = null
if (hostname == 'localhost') {
  keycloakRedirect = 'ffiec.cfpb.gov'
} else {
  keycloakRedirect = hostname
}

const keycloakConfig = {
  "realm": "hmda2",
  "url": `https://${keycloakRedirect}/auth`,
  "clientId": "hmda2-api",
  "public-client": true,
  "use-resource-role-mappings": true,
  "confidential-port": 0,
  "ssl-required": "all"
}```

PR #2515, opened 2025-04-30
feat(metrics): add custom event for OLARFT

### LINKED ISSUE #2514: [Metrics] Set up custom analytics event
We'll want to test out custom analytics events while we still have EDA support from Jen.

The event will be something like:

```
    analyticsSendEvent({
      event: 'OLARFT Events',
      action: 'Successfully saved row',
      label: editingRow["Record Identifier"],
    });
```

billhimmelsbach:
🚀 Currently on Dev: `v3.2.9b` 🚀

## Changes

- add `@cfpb/cfpb-analytics` library (same that [consumerfinance.gov uses](https://github.com/cfpb/consumerfinance.gov/blob/main/package.json#L21))
- add a sample test event to OLARFT when a row is saved (same as we did [in SBL](https://github.com/cfpb/sbl-frontend/blob/7169f2123db2a04dbd2fa18a58945883017c2a1f/src/utils/useSignAndCertify.ts#L32-L36))

## Testing

1. Does the OLARFT page load normally on Dev?

Closes #2514

### DIFF
--- a/src/tools/larft/components/row-editor/Editing.jsx
+++ b/src/tools/larft/components/row-editor/Editing.jsx
@@ -11,7 +11,7 @@ import { isRowTS } from '../../utils/row'
 import { EditingActions } from './EditingActions'
 import { EditingParsed } from './EditingParsed'
 import { EditingPiped } from './EditingPiped'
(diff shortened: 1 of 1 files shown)


ojbravo (approved):
Approved.

PR #2614, opened 2025-09-23
feat: make S3 bucket URL configurable

contolini:
Removes hardcoded S3 links that were scattered throughout the site and replaces them with a variable in our JSON config files. I tried to match the app's design patterns by relying on React context, higher order components and prop drilling when appropriate (turns out different sections of the app use different patterns). Non-React utility and constants files import the local config files using `getDefaultConfig`.

I named the variable `fileServerDomain` because it'll be the domain name of whatever service we end up using to serve the files. We're still deciding how the proxy will work in front of the new S3 bucket and we won't use raw S3 URLs anymore.

See [#5119](https://GHE/HMDA-Operations/hmda-devops/issues/5119)

## Changes

- Adds `fileServerDomain` property to config files.

## Testing

1. Testing is tricky because things work great locally but when deploying to dev, the app tries to fetch the new config file from GitHub which doesn't yet exist on `master`.

### DIFF
--- a/src/data-publication/reports/CombinedMLAR.jsx
+++ b/src/data-publication/reports/CombinedMLAR.jsx
@@ -1,12 +1,12 @@
 import { useEffect, useState } from 'react'
-import { withAppContext } from '../../common/appContextHOC.jsx'
+import { getDefaultConfig } from '../../common/configUtils'
 import LoadingIcon from '../../common/LoadingIcon.jsx'
 import { humanFileSize } from '../../common/numberServices.js'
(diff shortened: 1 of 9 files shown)


contolini:
@billhimmelsbach I ended up [removing](https://github.com/cfpb/hmda-frontend/pull/2614/commits/a9515b035402f4c65aeadc386a57ae787c59bb3b) all the HOC stuff and now every component reads the file server URL directly from the local config using `getDefaultConfig`.

contolini:
@billhimmelsbach Alex and I have been unable to recreate the blank screen issue soooo chalk it up to stale cache?

PR #2621, opened 2025-10-02
[Dev] Fix disclosure reports on Dev

billhimmelsbach:
After dropping the Dev S3 buckets, we need to fix the disclosure reports on Dev to point to Prod for its data products.

## Changes
- `src/data-publication/reports/Report.jsx`
  - always set the s3 bucket to be prod

- `cypress/e2e/data-publication/DisclosureReports.spec.js`
  - remove the `isDev` conditionals to always be prod
  - changed 2023 to an `msaMd` that the filing API shared on both Dev and Prod

- some eslint fixes here and there

## Testing

1. Do the tests pass against Dev? Yes!
Screenshot 2025-10-01 at 4 56 31 PM

2. Do disclosure reports still work?


contolini (approved):
The best type of tests is passing tests. 🥇

PR #2727, opened 2026-04-07
chore(docs): update docs for news page content management

contolini:
Updates the change log documentation and moves announcement "links" into markdown.

See [#5487](https://github.local/HMDA-Operations/hmda-devops/issues/5487).

## Changes

- Updates the [instructions](https://github.com/cfpb/hmda-frontend/blob/b845eb035b5fb4e5cb73836d68926dbfd4cc3020/src/updates-notes/README.md) on how to edit the [News and Updates](https://ffiec.cfpb.gov/updates-notes).
- Removes the [`links`](https://github.com/cfpb/hmda-frontend/blob/d9ead36afdeb94e35adce5f020dd62bdad47ea2c/src/updates-notes/change-log.md?plain=1#L79-L83) property from changelog entries. This property was needed when [JSON](https://github.com/cfpb/hmda-frontend/blob/master/src/updates-notes/change-log-data.json#L64-L73) was used to manage the announcements but now that we use markdown, which natively supports the creation of lists, we no longer need to programmatically generate a list of links at the end of each news entry. Editors can manually add lists of links or just use inline links throughout their content like a typical blog post.

## Testing

1. Do the [instructions](https://github.com/cfpb/hmda-frontend/blob/b845eb035b5fb4e5cb73836d68926dbfd4cc3020/src/updates-notes/README.md) make sense? Anything I should add/change?
2. Does `/updates-notes` at the PR preview below look the same as prod?

## Notes

There's a minor change to the padding of unordered lists. There used to be some padding that was added to `links` lists. Now that we don't use the `links` property, the bullet points are a little closer to the text (which IMHO is good because the width of the page is quite narrow).

| before | after |
|--------|-------|
| image | image |

### DIFF
--- a/cypress/e2e/updates-notes/ChangeLog.spec.js
+++ b/cypress/e2e/updates-notes/ChangeLog.spec.js
@@ -8,11 +8,6 @@ const REALLY_LONG_MOCK_DATA = `
 date: 03/01/99
 type: announcement
 product: tools
-links:
-  - text: "Cool link"
-    url: "https://consumerfinance.gov"
-  - text: "Another cool link"
-    url: "https://consumerfinance.gov"
 ---
 This is a super long announcement and it is very cool.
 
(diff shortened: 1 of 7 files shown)


billhimmelsbach (approved):
This looks great! Put it up on staging with the remote fetching disabled just to make sure, looks good!

If you want to improve this with Chynna's feedback, we can add the changes to this PR or you can make a new follow-up PR.

PR #2742, opened 2026-04-15
e2e(mlar): update staging cases, test checked state

billhimmelsbach:
To get mlar tests going again on staging again, we'll want to explicitly set the test cases for the staging environment to the same cases as prod. We also fix the "with headers" checkbox test after we fixed [the behavior over here](https://github.com/cfpb/hmda-frontend/pull/2729).

## Changes

- cypress/e2e/data-publication/ModifiedLAR.spec.js 
  - sets staging environment to use the same test cases as prod
  - makes a changes `cy.get('#inclHeader').check('true')` to `cy.get('#inclHeader').should('be.checked')` since we no longer have to manually set the checkbox to true [after I fixed it over here](https://github.com/cfpb/hmda-frontend/pull/2729). Instead we just do a quick assertion to make sure the checkbox is checked.
- cypress/support/helpers.js
  - add a `isStaging` helper function 

## Testing

1. Do the mlar tests pass on staging now? Yes!
passes-on-staging

2. Are the unit tests passing? Yep!

Screenshot 2026-04-15 at 10 09 42 AM

### DIFF
--- a/cypress/e2e/data-publication/ModifiedLAR.spec.js
+++ b/cypress/e2e/data-publication/ModifiedLAR.spec.js
@@ -1,5 +1,5 @@
 import { onlyOn } from '@cypress/skip-test';
-import { isBeta, isDev } from '../../support/helpers';
+import { isBeta, isDev, isStaging } from '../../support/helpers';
 const { HOST, YEARS } = Cypress.env()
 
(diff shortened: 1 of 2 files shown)


PatrickGoRaft (approved):
LGTM.

PR #2747, opened 2026-04-27
feat(akamai): add 500 pages with and without apor files

billhimmelsbach:
With the recent outage last month, we wanted to commit two version of the 500 pages that we could serve via Akamai: one with access to APOR files and one without. We'll want both: if the APOR files get stale for whatever reason, we won't want users to be directed to them.

These files can be served by themselves: everything is inlined to make it easy for the Akamai to serve. All the content was previously approved during the last outage.

## Changes

- adds 500 pages with and without APOR links

## Testing

1. Do the HTML pages load? Yes!
2. Do they look ok? Yes?

On staging as `<staging url>/akamai/500-with-apor-files.html`
Screenshot 2026-04-29 at 8 25 15 AM

On staging as `<staging url>/akamai/500.html`
Screenshot 2026-04-29 at 8 25 24 AM

## Notes
- The link back to home via the logo won't work until it's served from akamai. (I left it there to make it a little easier for a user attempting to retry/reload the home page)


contolini (approved):
5️⃣ 0️⃣ 0️⃣

PR #2750, opened 2026-04-29
fix(cache): replace hardcoded asset preload URLs with source URLs

contolini:
Vite replaces the URLs at build time with hashed minified versions.

See https://vite.dev/guide/assets.html
See https://vite.dev/guide/#index-html-and-project-root

## Changes

- Replace hardcoded USWDS asset URLs with their source URLs.

## Testing

1. Run `yarn build` and `yarn start` locally and check out the generated `index.html` file in the `/dist` dir.

### DIFF
--- a/index.html
+++ b/index.html
@@ -31,8 +31,6 @@
       type="font/woff2"
       crossorigin="anonymous"
     />
-    <link rel="modulepreload" href="/src/common/uswds/js/uswds-init.min.js" />
-    <link rel="modulepreload" href="/src/common/uswds/js/uswds.min.js" />
     <!--
       Notice the use of  in the tags above.
       It will be replaced with the URL of the `public` folder during the build.
--- a/index.html
+++ b/index.html
@@ -33,13 +33,13 @@
     />
     <link
       rel="preload"
-      href="/assets/uswds-init.min-D1tzE0C5.js"
+      href="/src/common/uswds/js/uswds-init.min.js"
       as="script"
       crossorigin="anonymous"
     />
     <link
       rel="preload"
-      href="/assets/uswds.min-BjNqw5Kx.js"
+      href="/src/common/uswds/js/uswds.min.js"
       as="script"
       crossorigin="anonymous"
     />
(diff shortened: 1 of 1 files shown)


billhimmelsbach (commented):
Thanks for the fix! Also... I put it up on staging and found some weird behaviors maaaaybe?

Looks like uswds.min.js gets downloaded twice maybe?

### First Load

Image

### Second Load
Screenshot 2026-04-30 at 10 26 48 AM



The uswds.min.js that is served from the disk cache is yours, but then it's still getting imported again [over here in Header.jsx](https://github.com/cfpb/hmda-frontend/blob/6e792cc43ddd5f166820ef827633f1ad90e3bd77/src/common/Header.jsx#L36) which is the one served from the memory cache.

Also, I'm not sure if the `uswds-init.min.js` is actually be preloaded? Looks like it's being served from the memory cache from the Header.jsx import instead of from the preload. Here's what the preload looks like, looks like in the index:

Image

Mind taking a quick peek? We can always just remove the preloads and see if it hits our lighthouse scores still, but maybe see if it's a quick fix?

contolini:
Lol okay how do you feel about just deleting the two preloads? It feels so dirty hardcoding the hashed filename into the HTML file. On staging, their deletion bumps the homepage from a 95 lighthouse score to a 98 so hey that's a success?

| before | after |
|--------|-------|
| image | image |

I tried [`modulepreload`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/modulepreload) over in https://github.com/cfpb/hmda-frontend/pull/2750/changes/02f91f0dc54d0e01ffaf5afd2c499054166df3a6 but it had no effect due to the structure of our app.

billhimmelsbach:
Yeah, let's remove them for now! 🚀

contolini:
> Yeah, let's remove them for now! 🚀

K they dead https://github.com/cfpb/hmda-frontend/pull/2750/commits/84f9b5963da2e180b08167ade769ce24e6222890

billhimmelsbach (approved):
Looks good! Good riddance!

PR #2757, opened 2026-05-05
fix(dockerfile): bump curl, remove pcre bump

billhimmelsbach:
Right now on `nginx:alpine3.23`, `curl` has vulnerabilities that haven't made it into the newest release yet, at least in a way that prisma cloud is happy with.

It's a little tricky to update `curl` manually, but I think it's worth it for now (with a follow-up ticket GHE #5550 to remove this temporary patch down the road if we want to).

The update we need for `curl` is only in the alpine edge repository.

I'm also taking the moment to remove the `pcre2` temporary fix, since it's been updated in the latest version of the docker image.

## Changes
- updates `curl` from the alpine edge repository (and their peer deps)
- removes the old `pcre2` patch that is no longer needed (prisma now passes without it)

## Testing
- Does this resolve the vulnerabilities to prisma's satisfaction? Yes!

hmda-frontend

- Does it look good on staging? Yes!


contolini (approved):
It builds and stagin' lookin' good! 👍

PR #2758, opened 2026-05-06
feat(nav): make filing button in main nav auth-aware

contolini:
When a user visits `/filing`, check their auth state and send them to either the filing "home" screen (the login page) or the institutions screen that lists their institutions.

This logic was already in the app but was incomplete. Currently, `HomeContainer.jsx` checks the app's state for the presence of an [open ID connect value](https://github.com/cfpb/hmda-frontend/blob/ab613603125eea12da34906c1a72636562de57bf/src/filing/home/HomeContainer.jsx#L7), which would indicate that the user has authenticated. The user is sent to the login screen if that [oidc value is `null`](https://github.com/cfpb/hmda-frontend/blob/master/src/filing/home/HomeContainer.jsx#L10-L12). However, it's always `null` because there's no code anywhere in the codebase that defines it. There's a [`USER_FOUND`](https://github.com/cfpb/hmda-frontend/blob/master/src/filing/reducers/user.js#L29-L34) action type that defines it but it's not used anywhere. This PR dispatches that action at the appropriate times and includes the user's keycloak info as the payload.

## Changes

- Dispatch `USER_FOUND` (via `userFound` action creator) when the filing app checks for `keycloak.authenticated` and it's `true`.
- Dispatch `USER_SIGNED_OUT` (via `userSignedOut` action creator) when authentication is not found by the filing app.
- In the `HomeContainer` component is the user's oidc property set?
  - No? Render the home component (this is the login page with the login.gov button).
  - Yes? Render the institutions component.
- Replace the logo anchor link with a `<Link>` to keep the entire nav within react router.

## Testing

1. e2e tests should pass.
2. When you click `Filing` and you're not logged in, does it show the login.gov page? After logging in, try clicking around the site and then clicking the `Filing` nav item. Does it take you directly to your institutions screen?

## Notes

- Redux's state is stored in memory so if you leave the app (e.g. by clicking over to the documentation site or refreshing the page), the `oidc` value will be cleared and the user will have to click the login.gov button again when visiting the filing page.

### DIFF
--- a/src/common/constants/links.js
+++ b/src/common/constants/links.js
@@ -50,11 +50,3 @@ export const defaultLinks = [
     ]
   }
 ]
-
-export const updateFilingLink = (config, links) => {
-  return links.map(link => {
-    if(link.name !== 'Filing') return link
-    link.href = `/filing/${config.defaultDocsPeriod}/`
-    return link
-  })
-}
--- a/src/filing/utils/keycloak.js
+++ b/src/filing/utils/keycloak.js
@@ -1,10 +1,10 @@
 /* eslint no-restricted-globals: 0 */
-import { error } from './log.js'
-import { getStore } from './store.js'
+import * as AccessToken from '../../common/api/AccessToken.js'
+import { getKeycloak } from '../../common/api/Keycloak'
 import isRedirecting from '../actions/isRedirecting.js'
 import { USER_FOUND } from '../constants'
-import { getKeycloak } from '../../common/api/Keycloak'
-import * as AccessToken from '../../common/api/AccessToken.js'
+import { error } from './log.js'
+import { getStore } from './store.js'
 
 const login = (path) => {
   const keycloak = getKeycloak()
@@ -95,7 +95,7 @@ const logout = (queryString = '') => {
 
   const logoutUrl =
     `${keycloak.authServerUrl}/realms/${keycloak.realm}/protocol/openid-connect/logout` +
-    `?client_id=${keycloak.clientId}` + // Use client_id instead of id_token_hint
+    `?id_token_hint=${encodeURIComponent(keycloak.idToken)}` +
     `&post_logout_redirect_uri=${postLogoutRedirectUri}`
 
   // Perform logout and redirect
@@ -103,4 +103,5 @@ const logout = (queryString = '') => {
   window.location.href = logoutUrl
 }
 
-export { register, login, logout, refresh, forceRefreshToken }
+export { forceRefreshToken, login, logout, refresh, register }
+
--- a/src/filing/utils/keycloak.js
+++ b/src/filing/utils/keycloak.js
@@ -26,6 +26,7 @@ const refresh = () => {
     console.error('Keycloak is not initialized')
     return
   }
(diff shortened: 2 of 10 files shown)


src/common/constants/links.js
```diff
@@ -50,11 +50,3 @@ export const defaultLinks = [
...
   }
 ]
-
-export const updateFilingLink = (config, links) => {
-  return links.map(link => {
-    if(link.name !== 'Filing') return link
-    link.href = `/filing/${config.defaultDocsPeriod}/`
-    return link
-  })
-}
```
contolini, line 60: The `Filing` link in the main nav uses this helper function to set its href to `/filing/<current-filing-period>`. It's unnecessary because our router [redirects `/filing` to the current filing period](https://github.com/cfpb/hmda-frontend/blob/ab613603125eea12da34906c1a72636562de57bf/src/filing/index.jsx#L58-L62) so it's much cleaner to just have the nav link to `/filing`.

billhimmelsbach (commented):
Looking great!

HMDA's sign out flow is weird, and maybe we should just get rid of this whole confirmation page anyway and go straight to login.gov, but it does have this weird "Filing Home" link that still exhibits the old behavior?

Image

contolini:
> Looking great!
> 
> HMDA's sign out flow is weird, and maybe we should just get rid of this whole confirmation page anyway and go straight to login.gov, but it does have this weird "Filing Home" link that still exhibits the old behavior?

@billhimmelsbach I removed this interstitial page via https://github.com/cfpb/hmda-frontend/pull/2758/changes/577c636f92f5141c10d73ba9c1ce881cebbf07df per our chat.

src/filing/utils/keycloak.js
billhimmelsbach, line 30:
```suggestion
const store = getStore()

```

billhimmelsbach, line 61:
```suggestion
const store = getStore()

```

billhimmelsbach (changes requested):
Heyo! It's looking good, but there might be a bug here that causes some issues?

When you call `store.dispatch` in `refresh()` and `forceRefreshToken()`, `store` is actually not defined in that function's scope. The other functions that call it use `getStore()` first to access it. So if you're logged into the filing app and one of these functions triggers: it'll error, hit the catch block with `login()`, and then bring ya back to the filing home.

You can test it out by idling on the profile page for awhile (`refresh()`), or by updating your profile (`forceRefreshToken()`). You'll see `ReferenceError: store is not defined` in the console then be taken back to the filing home. I put in some code suggestions where I'd put the `getStore()` in.

contolini:
> You can test it out by idling on the profile page for awhile (`refresh()`), or by updating your profile (`forceRefreshToken()`). You'll see `ReferenceError: store is not defined` in the console then be taken back to the filing home. I put in some code suggestions where I'd put the `getStore()` in.

Yuuuuup nice catch! Accepting your code suggestions fixes things. Thanks!

billhimmelsbach (approved):
Looks good now! Thanks for the work on it!

PR #2775, opened 2026-06-01
feat(profile): display form submission errors to user when updating profile

contolini:
When a user saves their institution profile, they don't see any errors if there is a failure, including the recent login.gov issues. This PR adds an error notification if the attempt to submit the profile form fails.

See [#5544](https://github.local/HMDA-Operations/hmda-devops/issues/5544)

## Changes

- The profile form now throws an error if the fetch is not `ok` (not a status in the range 200-299).
- The error is caught and displayed in an `Alert` notification at the top of the form with the error status and body serialized and base64-encoded.

## Testing

1. Head over to staging and force a 500 response from our `users/` endpoint by submitting a name change that includes a non-alphanumeric character.
2. Confirm that you can decode the error message by passing it to `atob()` in the browser console.

## Screenshots

https://github.com/user-attachments/assets/13d45083-6be0-4885-9d7d-f1cb44a5c87e

## Notes

Should the encoded error message have different text? Maybe the URL that returned the bad status code should be included in the message?

### DIFF
--- a/src/filing/profile/CompleteProfile.jsx
+++ b/src/filing/profile/CompleteProfile.jsx
@@ -133,7 +133,7 @@ const CompleteProfile = (props) => {
         })
         .then(async () => {
           setDisplayNotification(true)
-          setProfileSaveError(false)
+          setProfileSaveError('')
           setSaving(false)
           await forceRefreshToken()
           let newToken = jwtDecode(AccessToken.get())
--- a/src/filing/profile/CompleteProfile.jsx
+++ b/src/filing/profile/CompleteProfile.jsx
@@ -2,6 +2,7 @@ import { useEffect, useState } from 'react'
 import { Link, Prompt, Redirect } from 'react-router-dom'
 import Heading from '../../common/Heading'
 import InputAndLabel from '../../common/InputAndLabel'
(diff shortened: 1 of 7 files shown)


src/filing/profile/CompleteProfile.jsx
```diff
@@ -116,16 +118,29 @@ const CompleteProfile = (props) => {
...
+            const saveError = new Error(
+              `Failed to update user profile (${response.status}: ${responseBody})`,
+            )
+            throw saveError
+          }
+          return response.json()
+        })
         .then(async () => {
           setDisplayNotification(true)
+          setProfileSaveError(false)
```
billhimmelsbach: Nit: Did you want this to be `setProfileSaveError('')` to match the way you initiated it as an empty string on line 39?

billhimmelsbach (approved):
Looks great! One small code comment and...

> 1. Should the encoded error message have different text? Maybe the URL that returned the bad status code should be included in the message?

I think this is ok for now? It's a good start we can iterate on, but maybe we should create a broader error handling ticket that we can use to lift this error handling work into a component that can be used elsewhere / potentially improve the error handling with a URL, etc?

Good work Tommy Pickles! 👶

PR #2779, opened 2026-06-04
fix(profile): improve UX/a11y of profile page save functionality

contolini:
Improvements to the filing app profile page save functionality.

See [#5543](https://github.local/HMDA-Operations/hmda-devops/issues/5543)

## Changes

- Updates the text color of our disabled buttons (throughout the site) to make them 508-compliant. They now use `#454545` which is what USWDS recommends.
- Adds a new `SpinnerDots` component with CSS copied from [login.gov](https://github.com/18F/identity-idp/blob/46ec27d430f90d7b128d4e6be57d34478f8002a3/app/assets/stylesheets/components/_spinner-dots.scss)
- Moves the missing institutions banner out of the `Alert` component to prevent banner blindness.

## Testing

1. PR tests should pass.
2. Head over to staging and update your profile and see how it looks.

## Notes

- The missing institutions banner also appears at the bottom of the [institutions landing page](https://github.com/cfpb/hmda-frontend/blob/6bd0e090d98a0ca8f7523e50aab9195ac19edb2e/src/filing/institutions/index.jsx#L243) so this change will also change its appearance on that page. IMHO this is good because there are already multiple alerts on that page and the goal is to avoid banner blindness.

## Screenshots

https://github.com/user-attachments/assets/251a30c3-12fa-4b39-9a46-e7e04e5eced5

### DIFF
--- a/cypress/e2e/filing/Profile.spec.js
+++ b/cypress/e2e/filing/Profile.spec.js
@@ -102,5 +102,24 @@ describe(
       cy.get('.profile_error_code_box').should('be.visible')
       cy.get('.copy_profile_error_button').contains('Copy error to clipboard')
     })
+
+    it('Shows spinner dots in save button during save', () => {
+      cy.visit(`${AUTH_BASE_URL}filing/profile`)
+      cy.wait(ACTION_DELAY)
+
+      cy.get('.profile_form_container > :nth-child(2) > input').clear().type('party')
+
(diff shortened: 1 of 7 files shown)


billhimmelsbach (approved):
Looks good!

I agree that the change to the bottom of the filing home is an improvement over the banner pile before.

PR #2782, opened 2026-06-14
fix(linter): fix config and deps causing lint to fail

billhimmelsbach:
Couple issues were breaking linting:
- `languageOptions` in `.eslintrc.json` isn't supported by the eslint version we're on, `8.x`. That's not included till `9.x`.
- We had a mismatched `minimatch` versions that resulted from our dependency resolution (that I think is no longer needed for security).

I thought about being more ambitious with this fix and getting prettier/eslint working a little better in a github action, but decided to keep things simple until after data publication season.

I did at least include a rule for json/jsonc files, so that they always end in a newline character on save at least for vscode (rather than trying to dig through prettier vs eslint stuff for now)

## Changes

- .vscode/settings.json
  - ensure there's always one newline char at the end of .json files
-  .eslintrc.json
  - remove nested `languageOptions`
- package.json
  - remove conflicting `minimatch` dependency

## Testing

1. After you `yarn install`, can you save and see linting changes? Try looking at `src/MaintenanceMessage.jsx` and see if on save the excess React import at the top of the file disappears (while the un-autofixable `explicit type attribute` error stays)

## Notes

- I've got a [follow-up PR](https://github.com/cfpb/hmda-frontend/pull/2783) to do the auto-fixable linting issues all at once.


contolini (approved):
❤️

PR #2783, opened 2026-06-14
fix(linter): run linter with new settings and fixes

billhimmelsbach:
## Changes

- runs the linter rules after the fixes from https://github.com/cfpb/hmda-frontend/pull/2782

## Testing

1. Do the linting fixes look reasonable at a glance?
2. Do the tests still pass?

## Notes
- I have a fix ENT: #5580 for batching all the sorting of imports too, but I'll save that for later since it hypothetically might break things with css, etc. I'll save it till after data publication.


contolini (approved):
❣️

PR #2784, opened 2026-06-16
fix(data): add loading state to MSA/MD fetch

contolini:
We use `react-select` for the MSA/MD typeahead which has an `isLoading` prop but we've never used it. We now pass `true` when we're waiting for the MSA response from the server.

I also had to bump `react-select` to v5 to make it compatible with React 18+.

Fixes [#5323](https://github.local/HMDA-Operations/hmda-devops/issues/5323)
See https://react-select.com/props
See https://github.com/JedWatson/react-select/pull/5689

## Changes

- Replaced the broken loading state logic with `react-select` loading state logic.

## Testing

1. PR tests should pass.
1. It's not currently on staging due to ongoing data publication testing stuff but check it out locally.

## Notes

There was already a loading state built into the app but `isLoaded` always returned truthy because we're [double banging an empty array](https://github.com/cfpb/hmda-frontend/blob/7846bb89942d6a7f1c8555e8fcb6956336f04e17/src/data-publication/reports/MsaMds.jsx#L16-L17) and !![] = true in JS. I could just fix that logic but IMHO it's much better UX to use the native `react-select` loading state.

## Screenshots

https://github.com/user-attachments/assets/03ad3f54-d8e8-4beb-be0d-d8c8c495e9b0

https://github.com/user-attachments/assets/4b530581-0b47-4a0f-b606-295c86f0768e

### DIFF
--- a/cypress/e2e/data-publication/AggregateReports.spec.js
+++ b/cypress/e2e/data-publication/AggregateReports.spec.js
@@ -18,9 +18,9 @@ onlyOn(!isBeta(HOST), () => {
       cy.visit(`${HOST}/data-publication/aggregate-reports/2024`)
 
       // Select geography and report
-      cy.findByText('Select a state...').type('Arizona{enter}')
-      cy.findByText('Select MSA/MD...').type('Phoenix{enter}')
-      cy.findByText('Select report...').type(
+      cy.findByRole('combobox').type('Arizona{enter}')
+      cy.findByRole('combobox').type('Phoenix{enter}')
+      cy.findByRole('combobox').type(
         'Applications by Ethnicity and Sex{enter}',
       )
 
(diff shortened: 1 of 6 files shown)


billhimmelsbach:
Excited for this one @contolini! I think some tests might still be failing though?

contolini:
@billhimmelsbach I changed the cypress selector to find the `react-select` element by using its [a11y role instead of placeholder text](https://github.com/cfpb/hmda-frontend/pull/2784/commits/9c5d41201278ae58cd3139a84d9e06e78ead9c8a) to get the tests back online. And the `Selector` component that I updated is only used in two files, [`MsaMds.jsx`](https://github.com/cfpb/hmda-frontend/blob/master/src/data-publication/reports/MsaMds.jsx) and [`Reports.jsx`](https://github.com/cfpb/hmda-frontend/blob/master/src/data-publication/reports/Reports.jsx), and both those files are covered by the updated e2e tests.

billhimmelsbach (approved):
Great! 🎉

I went through and checked the implementations of react-select, and I think the only styling regression that's noticeable is on hmda-help where it looks like the text isn't centered anymore? Doesn't seem like a biggie since it's not for the user anyway?

Image

So it gets a 👍 from me!

PR #2785, opened 2026-06-17
Data publication draft 2025

billhimmelsbach:
See https://github.com/cfpb/hmda-frontend/pull/2791 for the final version that was deployed


src/data-browser/maps/MapContainer.jsx
```diff
@@ -454,6 +460,21 @@ const MapContainer = (props) => {
...
 
+  useEffect(() => {
+    if (
+      !county2025Data &&
+      selectedGeography.value === "county" &&
+      year === "2025"
+    ) {
+      fetchQ.push(1)
+      runFetch("/2025/county.json").then(jsonData => {
+        console.log(jsonData)
```
contolini:
```suggestion
```

This `console.log` was copied over from the 2024 code block above it and should prob be deleted from there as well.

contolini: Also there are 26 `useEffect`s in this file. In a distant future we should question why that is. Re: https://react.dev/learn/you-might-not-need-an-effect

billhimmelsbach: Ha! I hear you. The useEffect pile is wild, but it'll be what I do this year.

I can kill that console.log!

contolini (approved):
🚀

billhimmelsbach:
See https://github.com/cfpb/hmda-frontend/pull/2791 for the final version that was deployed

PR #2790, opened 2026-06-18
feat(reports): show notice when all values are zero

billhimmelsbach:
This one was fun: when a report's table is only filled with zeroes, we wanted to be able to show a notice to the user instead of having them hunt for a non-zero value. I think this is a pretty clean frontend-only solution. Definitely worth a little sleep deprivation.

## Changes

- Report.css
  - adds a little CSS for the notice (I generally avoid negative margins, but the way the table is laid out made this a cleaner solution)
- Report.jsx
  - adds hasNonZeroValue: a cute recursive function to dig through and exit early if it finds a non-zero value for keys containing `count` or `value` which covers the different reports we have
  - adds areAllMetricValuesZero: change the state of `hasNonZeroValue` if a non-zero value is found, otherwise on error don't display anything. Skip IRS reports since they always have something and have a different format.
  - adds the notice to the table header: there are better ways of doing tables for accessibility, but that's for another day 

## Testing

1. Check out a report that only contains zeros, try `[STAGING]/data-publication/disclosure-reports/2024/5493007KLU19FK8PG279/38060/2` and see if the notice appears
2. Check out a report that has data, try `[STAGING]/data-publication/disclosure-reports/2024/5493007KLU19FK8PG279/11694/1`
3. Are the new unit tests passing? Yes!
    Screenshot 2026-06-21 at 12 24 53 PM

4. Are the new e2e tests passing against staging? Yes!
    Screenshot 2026-06-21 at 12 25 00 PM

## Notes

- Having now dug through these tables, there's a lot of helpful accessibility improvements that could be done on them someday.

## Screenshots
Screenshot 2026-06-18 at 1 55 03 AM

### DIFF
--- a/src/data-publication/reports/reportUtils.js
+++ b/src/data-publication/reports/reportUtils.js
@@ -1,0 +1,34 @@
+/**
+ * A cute recursive function to dig through and exit early if it finds a non-zero
+ * value for keys containing "count" or "value"
+ * @param {*} node - The report object or nested object to search for non-zero values
+ * @returns {boolean} True if a non-zero value is found, otherwise false
+ */
+export const hasNonZeroValue = (node) => {
+  if (!node || typeof node !== 'object') return false
+  return Object.entries(node).some(([key, value]) =>
+    /(count|value)$/i.test(key)
+      ? typeof value === 'number' && value !== 0
+      : hasNonZeroValue(value),
+  )
+}
+
+/**
+ * Runs hasNonZeroValue for any given report except those without numerical values like the
+ * IRS and Reporting Financial Institutions ("I") report
+ * @param {object} report - A report object to check for non-zero values
+ * @returns {boolean} True if every value is zero, otherwise false
+ */
+export const areAllReportValuesZero = (report) => {
+  if (report.table === 'IRSCSV') return false
+  if (report.table === 'I') return false
+  try {
+    return !hasNonZeroValue(report)
+  } catch (error) {
+    // not sure what to do other than log it for a poor dev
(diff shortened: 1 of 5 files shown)


src/data-publication/reports/reportUtils.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * A cute recursive function to dig through and exit early if it finds a non-zero
+ * value for keys containing "count" or "value"
+ * @param {*} node - The report object or nested object to search for non-zero values
+ * @returns {boolean} True if a non-zero value is found, otherwise false
+ */
+export const hasNonZeroValue = (node) => {
+  if (!node || typeof node !== 'object') return false
+  return Object.entries(node).some(([key, value]) =>
+    /(count|value)$/i.test(key)
```
contolini, line 12: Are we looking for a key that ends in `count` or `value` or are we looking for keys that equal `count` or `value`? If the latter, I'd do something like `key === 'count' || key === 'value'` here.

contolini (approved):
Works great! One minor comment.

src/data-publication/reports/reportUtils.js
billhimmelsbach: Heyo! Yeah you found some of my pain @contolini. There are some sneaky ones that only end in count or value like `noLienCount` or `firstLienValue` that would get missed by `key === 'count' || key === 'value'`.

We also have to keep the "ends with" because there are some keys like `values` with just arrays in them that we don't want to grab.

I can add a comment to make that clearer.

PR #2792, opened 2026-06-23
fix(homepage): update youtube modal css to allow play

billhimmelsbach:
Let's fix that YouTube modal. If you check it out on prod by clicking "See how it works," you can't click play on the video without it closing, which is a bummer. Looks like it was broken on a uswds update a while ago, but since its css only affects this one modal, I just tinkered with it until it looked better. There's more that could be done in the future, but this is a heck of a lot better for now.

## Changes

- updates the uswds override CSS to allow the YouTube video to play and be closed again
  - stops the close button from taking over the whole div and preventing clicks
  - repositions the close button farther into the upper right corner of the video player so that it's easier to close without hitting the other controls
  - makes the close button bigger so that it's easier to click
- remove a button class on the close button that was causing some weirdness when trying to close the modal
- update the compiled uswds CSS (automated)

## Testing

1. How does it look on staging? Good! (tagged on staging as `staging-data-publication-rc-2`)
2. Do the tests still pass? Yes!
Screenshot 2026-06-23 at 4 25 04 AM

## Notes

- More could be done, but this is a lot better than what we have now. For example, I'd like to add in a "clicking away from the modal, closes the modal" in the future.

## Screenshots

On staging now:

Screenshot 2026-06-23 at 4 26 15 AM


PatrickGoRaft (approved):
LGTM

PR #2794, opened 2026-06-23
fix(announcements): double time between autoadvances

billhimmelsbach:
## Changes

- just makes the time longer between the announcement carousel advancements so that people have a chance to read it a bit more (5 seconds to 10 seconds)

## Testing

1. How does it look on staging? Good! (tagged on staging as `staging-data-publication-rc-2`)
2. Do the tests still pass? Yes!
Screenshot 2026-06-23 at 4 29 54 AM

## Notes

- We can tune this value more in the future, but this is a good start.

## Screenshots
Screenshot 2026-06-23 at 4 27 01 AM


PatrickGoRaft (approved):
LGTM

PR #2795, opened 2026-06-23
test(profile): remove redirect start to fix loading test

billhimmelsbach:
## Changes

- the loading test was failing for me against staging because the initial redirect to the profile page via url only redirected the test to filing home

## Testing

1. Does the test now work against staging? Yes!
Screenshot 2026-06-23 at 6 33 05 AM

## Notes
- if the intention was to keep the tests isolated, maybe we can click the profile link between tests?


contolini (approved):
Mmmm yeah my goal was to isolate the tests but that should be done in a `beforeEach` anyway

PR #2074, opened 2024-01-26
[Vite] Refactor React code

Michaeldremy:
Closes #1909 

There are additional items included in this PR:

- Updated Disclosure and MLAR Cypress tests to be more specific on the institution it's trying to search for
- Added Jest configuration to run the old Jest/Enzyme unit tests
- Added Configuration + Setup to start creating Cypress Component tests

PR #2800, opened 2026-06-26
feat(msaMds): update lists based on source of truth JSONs instead of vars

billhimmelsbach:
We received a report that aggregate reports were missing some MSA/MDs, specifically missing a number of MSA/MDs in Montana such as Bozeman and Helena. These values were drawn from this file [data-publication/constants/stateToMsas.js](https://github.com/cfpb/hmda-frontend/blob/0c2f7b08f4f0b32fa04ea776cd95b8ea8f3cbbec/src/data-publication/constants/stateToMsas.js).

It looks like the values haven't been consistently updated (or correctly updated) for the last ~6 years for aggregate reports . Also for the data browser, the data is inconsistent with the official MSA/MD files as well. These values are drawn from [msaToName.js](https://github.com/cfpb/hmda-frontend/blob/0c2f7b08f4f0b32fa04ea776cd95b8ea8f3cbbec/src/data-browser/constants/msaToName.js), [msaToState.js](https://github.com/cfpb/hmda-frontend/blob/0c2f7b08f4f0b32fa04ea776cd95b8ea8f3cbbec/src/data-browser/constants/msaToState.js), and [stateToMsas](https://github.com/cfpb/hmda-frontend/blob/0c2f7b08f4f0b32fa04ea776cd95b8ea8f3cbbec/src/data-browser/constants/stateToMsas.js) in the data browser folder. Now that I've run the underlying frontend data publication scripts that the backend usually runs, this makes some sense, since the variables in these files are also hard to update and have to be done by hand (and are updated in different ways).

Since the frontend started using the frontend data scripts (see repo: automated_analytics), we already generate the json data we need ourselves and can just plop each new year directly into the frontend. Instead of hand crafting the new variables, they can just be added for the snapshot, one year, and three year each data publication season. These files are super small: worth the few extra kilobytes for not having to parse through MSA/MD updates again.

There are some caveats discussed in the Caveat Corner section below, but generally, this aligns the frontend with the backend and accurate MSA/MDs generally. It's also is way easier and easier to maintain and document.

## Changes

- add the msaToName, msaToState, and stateToMsas files for 2017 to 2025 from the frontend data publication scripts (auto generated)
- update in the data-browser/constants folder: msaToName, msaToState, and stateToMsas files to just use the JSON files instead of hand crafted variables
- update in the data-publication/constants folder: change stateToMsas.js to labeledMsaMdsByYearAndState.js to generate a file based on msaToName and stateToMsas data to match what used to be handled by stateToMsas.js for aggregate reports. Adds a little function to generate the expected format instead of using variables
- update Aggregate.jsx and MsaMds.jsx to use the new labeledMsaMdsByYearAndState file

## Caveat Corner
_See Caveat Corner Update below to see the solution to most of the issues below_

So the only bummer about this approach is that the names and associated states for MSA/MDs for 2019 and 2020 are wrong based on the source of truth files, 2019_public_msamd_three_year_csv.csv and 2020_public_msamd_three_year_csv.csv that the frontend data publication scripts use.
While working on this, I noticed something weird: for the [2019_public_msamd_three_year_csv.zip](https://files.ffiec.cfpb.gov/static-data/three-year/2019/2019_public_msamd_three_year_csv.zip) and [2020_public_msamd_three_year_csv.zip](https://files.ffiec.cfpb.gov/static-data/three-year/2020/2020_public_msamd_three_year_csv.zip) the MSA/MDs listed aren't always MSA/MDs but are often maybe mislabeled as the Combined Statistical Area or CSA that they are a part of.
 
For example, usually the MSA/MD 10420 is associated with AKRON, OH. However, in 2019 and 2020, this MSA/MD 10420 is maybe mislabeled as Cleveland-Akron-Cantonwhich is the CSA not the MSA/MD as I understand? I think this is possible true for many of the MSA/MDs in the 2019 and 2020 files, which causes some weird data issues.
 
This is the same problem as disclosure reports too on prod, which uses the API instead of a static file. Check it [out here](https://ffiec.cfpb.gov/data-publication/disclosure-reports/2019/KB1H1DSPRFMYMCUFXT09) and type in "Cleveland" (you might have to wait a bit before the page loads). You'll see three MSA/MDs with the same name, Cleveland-Akron-Cantonwhere all three are labeled with the CSA instead of the MSA/MD.

The data is fine for each of these MSA/MDs if you click into them, but it's confusing that they are mislabelled (and sometimes in multiple states). So it's not a data quality issue per se, but it makes it sometimes harder to find the MDA/MD you want.
 
The backend/data team are aware of this issue, and they may end up fixing it in the future. For now, at least this change makes us align with the source of truth / backend, even if we're now identically inaccurate for 2019/2020 for some MSA/MDs names and their associated states.

Check out the conversation in HMDA Dev for more details.

## Caveat Corner Update

So I dug in because the 2019/2020 issues made me sad. I took the 2019/2020 data with its CSAs instead of MSA/MDs and started correcting them to use the right labels and associated states for the given MSA/MD codes. When I started doing it and double checking the other jsons to make sure I was doing it right, I noticed that the values for 2019 matched the 2021 data.

Looking through the [OMB bulletins](https://www.census.gov/programs-surveys/metro-micro/about/omb-bulletins.html), which is how these MSA/MD changes are announced, I realized that there were OMB Bulletins that were announced in 2018 (which affects 2019 data), but not again until 2023 (which affected 2024 data). There was a bulletin released in March 2020, but it made changes only to micropolitan areas, not the MSA/MDs we use. 

So if there haven't been changes between 2019, 2020, 2021, 2022, and 2023, we can use data from 2021 for both 2019 and and 2020.

Testing this out against the backend (which gives the right data from the API when you give it the year and MSA/MDs codes), confirms that everything is working great. Check out these URLs for 2019/2020 data that's uses MSA/MDs that didn't exist in the 2018 data but are the same as 2021 data. Since these urls resolve, that's a sign we're good to go:

**Aggregate Report for new MSA/MD code for Chicago for 2019 and 2020:**
`[STAGING]/data-publication/aggregate-reports/2019/IL/16984/1`
`[STAGING]/data-publication/aggregate-reports/2020/IL/16984/1`

**Data Browser for new MSA/MD code for Chicago for 2019 and 2020 (might take a bit to load):**
`[STAGING]/data-browser/data/2019?category=msamds&items=16984&leis=549300MIDG4TRYO5ON35&loan_types=1,2,3,4&getDetails=1` 
`[STAGING]/data-browser/data/2020?category=msamds&items=16984&leis=549300MIDG4TRYO5ON35&loan_types=1,2,3,4&getDetails=1` 

**Aggregate report for new MSA/MD code for Yauco, Puerto Rico for 2019 and 2020**
`[STAGING]/data-publication/aggregate-reports/2019/PR/49500/1`
`[STAGING]/data-publication/aggregate-reports/2020/PR/49500/1`

**Data Browser for new MSA/MD code for Yauco, PR for 2019 and 2020 (might take a bit to load):**
`[STAGING]/data-browser/data/2019?category=msamds&items=49500&leis=OJH4M4RQMT1WXN2BNC32&loan_types=1,2,3,4&getDetails=1`
`[STAGING]/data-browser/data/2020?category=msamds&items=49500&leis=OJH4M4RQMT1WXN2BNC32&loan_types=1,2,3,4&getDetails=1`

This doesn't fix the wonky backend API calls that Disclosure Reports use for 2019/2020, but at least we've done what we can on the frontend. In the future, I'd just use the frontend and bypass the backend entirely to save the slow calls against disclosure reports.

## Testing

1. Do the tests still pass! Yes!
2. Do you see 5 MSA/MDs for aggregate reports for Montana on staging? Yes! (tagged as `5661-and-5671-security-and-aggregate-reports`)
3. Do aggregate reports and the data browser look good on staging? Yes!

## Notes
- I found another unrelated bug in aggregate reports on prod where trying to get to an aggregate report directly via URL instead of through the UI can cause a crash (check out [this URL](https://hmda-eks.staging-regtech.aws.cfpb.gov/data-publication/aggregate-reports/2025/AK/11260/1)). I have another PR I'll put up for that after we merge this one in.
- Since some of these JSONs are the same, you could imagine that we could reuse the JSON from other years. I'd rather just add a few kilobytes each year unless it becomes a problem since re-using got us into this mess in the first place?
- I'm sorry for this book I've written Chris.

Closes GHE #5661


src/data-publication/reports/MsaMds.jsx
```diff
@@ -1,5 +1,5 @@
 import { useEffect, useState } from 'react'
-import stateToMsas from '../constants/stateToMsas.js'
+import labeledMsaMdsByYearAndState from '../constants/labeledMsaMdsByYearAndState.js'
```
contolini, line 2: Does [detailsCache.jsx](https://github.com/cfpb/hmda-frontend/blob/4c6f45536cf807860f35768d25b5c58909dfa98b/src/data-publication/reports/detailsCache.jsx#L10) need the same treatment? (swapping out `stateToMsas` for `labeledMsaMdsByYearAndState`)?

billhimmelsbach: A great question that I spent way too long on investigating! I started with updating that file too but... I don't think it's actually used anywhere?

I tried searching for any file that imported it (no luck), then I ran this silly tool call [madge](https://github.com/pahen/madge) to see if there were any weird imports of it, then I just tried deleting it and seeing if the build still worked (it did).

Soooo... I think it's ok to leave alone @contolini? I could leave a comment at the top, but I think I kind of want to just go through in a separate PR and mark all the dead code I found with madge?

contolini: > I think I kind of want to just go through in a separate PR and mark all the dead code I found with madge?

Sounds good!

contolini: It looks like that file was created in [#2074](https://github.com/cfpb/hmda-frontend/pull/2074/changes#diff-9382c7fc7875ef7ccaa22ac5b67233be2fdab6393274a718796a4d6df28b7715) but never actually used anywhere. It was probably supposed to be imported into [`NationalAggregate.jsx`](https://github.com/cfpb/hmda-frontend/blob/87924edbc5beab792cc7751f2e3a1bff9912988c/src/data-publication/reports/NationalAggregate.jsx#L13) but the PR forgot to wire it in.

billhimmelsbach: Added a ticket GHE #5681 to track it! 🎉

contolini (approved):

PR #2802, opened 2026-07-01
fix(security): bump dompurify, form-data, js-yaml

billhimmelsbach:
Bumps a few dependencies to patch a few vulnerabilities.

## Changes

Resolution changes:
- update `dompurify` `3.4.5` to `>=3.4.11`
- update `form-data` `>=4.0.4` to `>=4.0.6`

New resolutions:
- add `js-yaml@3` at `>=3.15.0`
- add `js-yaml@4` at `>=4.2.0`
- add very specific vite resolution to target only a vulnerable version used in `vitest` at  `vite@^5.0.0 || ^6.0.0 || ^7.0.0-0` at `^7.3.6`

## Testing

1. Do the test still pass? Yes!
2. How do things look on staging? Good! (on staging as `5661-and-5671-security-and-aggregate-reports`)

## Notes
- I think we should switch things over to `^` rather than the norm of `>=` for resolutions. It was kind of an anomaly in some of the hmda-frontend resolutions, and I don't want to keep adding to the weirdness. There's some weirdness that can happen with `>=` which can cause issues if one of the resolutions had to be re-resolved or we ever had to regenerate the lock file. So I've added a bullet to this existing resolution cleanup issue GHE #5579, and we can clear these security issues first and put it in the TODO column for next sprint.

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,13 +140,13 @@
     "tough-cookie": ">=4.1.3",
     "word-wrap": ">=1.2.4",
     "browserify-sign": ">=4.2.2",
-    "dompurify": ">=3.4.11",
+    "dompurify": "^3.4.11",
     "esbuild": "0.25.0",
     "canvg": "3.0.11",
     "elliptic": "6.6.1",
     "cross-spawn": ">=7.0.6",
     "glob": ">=11.1.0",
-    "form-data": ">=4.0.6",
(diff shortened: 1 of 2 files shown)


contolini (approved):
> I think we should switch things over to ^ rather than the norm of >= for resolutions.

Agreed 👍

PR #2803, opened 2026-07-01
fix: link only to 2017 national aggregate reports

contolini:
The national aggregrate reports data product was only generated in 2017, not any of the other years, so the homepage link should not be dynamically generated with the current data publication year.

See [#5667](https://github.local/HMDA-Operations/hmda-devops/issues/5667)

## Changes

- Hardcode NAR link to 2017

## Testing

1. I have not deployed this to staging because it's... fairly straightforward.

## Notes

- I was unable to find any other links to national aggregrate reports so lmk if I missed anything.


billhimmelsbach (approved):
Looks good to me!

PR #2804, opened 2026-07-02
fix(prod-config): adjust filing dates to match the fig

billhimmelsbach:
We'll want to update the prod-config to match the dates investigated in GHE #5375 after I saw some drift in the dates when preparing for filing season.

## Changes

- matches exactly what is listed in GHE #5375

## Notes

- We'll want to restore the quarterly filing schedule on the filing home after this PR later:
https://github.com/cfpb/hmda-frontend/blob/5375-update-filing-dates/src/filing/common/FilingPeriodsCard.jsx#L62-L72


contolini (approved):
Great catch!

PR #2807, opened 2026-07-08
fix: open FIG edit links in new tab

contolini:
## Changes

- Links to the FIG now open in a new tab.

## Notes

- I added `noopener noreferrer` to appease cyber scanners

fig-links-after


billhimmelsbach (approved):
Looks good!

PR #2815, opened 2026-07-22
feat(filing): add markdown support to filing descriptions

billhimmelsbach:
Let's add markdown for edit descriptions! This PR adds basic markdown formatting particularly necessary for links, with the potential to add support for further markdown features later.

## Changes
_src/filing/submission/edits/Table.css_
- adds CSS for the `markdown-description` wrapper to mimic the existing two rules [here](https://github.com/cfpb/hmda-frontend/pull/2815/changes#diff-67d473742f7a5f19711d8ec0fbc7684c32f91d6f8a8776b0bf87e7773e783851R39-R40) and [here](https://github.com/cfpb/hmda-frontend/pull/2815/changes#diff-67d473742f7a5f19711d8ec0fbc7684c32f91d6f8a8776b0bf87e7773e783851R78) to avoid everything being bolded or the wrong color

_src/filing/submission/edits/Table.jsx_
- lints [these lines](https://github.com/cfpb/hmda-frontend/pull/2815/changes#diff-f0418b333d12607045516ff3ad5fe9969a70b43e246c3c5bd8e811ee84e0f551R70-R76) with no substantive changes
- strip out only the beginning and ending quotes, instead of all quotes
- render the edits with markdown using `rehype-external-links` to open in a new tab
- remove the summary property that [was deprecated awhile ago](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTableElement/summary) in favor of just using the caption which we already do and would render weird markdown now

_package.json_
- adds `rehype-external-links` so we can open links in edit descriptions in new tabs

## Testing
- How does it look on staging? Good! (currently on staging as `2806-add-markdown-parsing-for-filing-app-1`) I rendered every unique description since 2018 and compared it to the new markdown version. Looks good! There are a few edits that are a little spicy like Q660-3 that includes `*`, it's separated by commas so doesn't trigger any markdown styling.

Screenshot 2026-09-17 at 3 22 33 PM

We'll want to test out any new edits with this new markdown support.

I ran some image diffing on the before and after the markdown changes. You can view every one of the edits rendered in [this pdf](https://github.com/user-attachments/files/32358232/MD.Test.-.All.Markdown.Edits.Since.2018.pdf) if you want to take a look. You can also deploy it to staging again if you want to, it's tagged as `markdown-support-demo-2`

- Do the tests still pass? Yes!

### DIFF
--- a/src/filing/submission/edits/Table.jsx
+++ b/src/filing/submission/edits/Table.jsx
@@ -96,13 +96,13 @@ export const renderTableCaption = (props) => {
     captionHeader = 'Review your loan/application IDs'
   }
 
-  const { description } = props.edit
+  const cleanDescription = props.edit.description.replace(/^"|"$/g, '')
 
   if (shouldSuppressTable(props)) {
     return (
       <div className='caption'>
         <h3>{captionHeader}</h3>
-        {renderDescription(description)}
+        {renderDescription(cleanDescription)}
         {name === 'S040' ? (
           <p>
             Please check your file or system of record for duplicate
@@ -116,7 +116,7 @@ export const renderTableCaption = (props) => {
   return (
     <caption>
       <h3>{captionHeader}</h3>
-      {renderDescription(description)}
+      {renderDescription(cleanDescription)}
     </caption>
   )
 }
--- a/src/filing/submission/edits/Table.jsx
+++ b/src/filing/submission/edits/Table.jsx
@@ -1,4 +1,6 @@
 import PropTypes from 'prop-types'
+import Markdown from 'react-markdown'
+import rehypeExternalLinks from 'rehype-external-links'
 import Loading from '../../../common/LoadingIcon.jsx'
 import { splitEditPart, splitYearQuarter } from '../../api/utils.js'
 import Pagination from '../../pagination/container.jsx'
@@ -65,7 +67,13 @@ export const renderTableCaption = (props) => {
   const [edit] = splitEditPart(name)
 
   const linkedName = (
-    <a href={`/documentation/fig/${year}/overview#edit-${edit}`} target='_blank' rel='noopener noreferrer'>{name}</a>
+    <a
+      href={`/documentation/fig/${year}/overview#edit-${edit}`}
+      target='_blank'
+      rel='noopener noreferrer'
+    >
+      {name}
+    </a>
   )
   let captionHeader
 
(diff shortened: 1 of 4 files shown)


src/filing/submission/edits/Table.jsx
```diff
@@ -88,13 +96,13 @@ export const renderTableCaption = (props) => {
     captionHeader = 'Review your loan/application IDs'
   }
 
-  const description = props.edit.description.replace(/"/g, '')
+  const { description } = props.edit
```
contolini: The backend wraps all the descriptions in double quotes so they should prob be trimmed from the start and end of the descriptions, something like:

```suggestion
  const { description } = props.edit.description.replace(/^"|"$/g, '');
```

billhimmelsbach: Heyo! Sounds good to me @contolini!

I don't thiiink that kind of assignment would work though since we're destructuring `props.edit` and pulling out `description`, but `props.edit.description.replace(/^"|"$/g, '')` returns a string so we'd get `undefined` / nothing would render if we tried to destructure a `description` property from it.

How about something descriptive like:
`const cleanDescription = props.edit.description.replace(/^"|"$/g, '')`

And then changing `description` to `cleanDescription` where we use it?

contolini: Ah yeah that works. `description` like in the original code is also fine by me. Whichever you prefer.

billhimmelsbach: @contolini

I ran a janky visual regression test against every unique edit description from 2018 to 2026: the old renders versus the new markdown renders (you can see all the edits with and without markdown on staging currently at `[STAGING]/filing/2026/B90YWS6AFX2LGWOXJ1LD/syntacticalvalidity`)

I didn't find any breaking changes which is sweet: the only thing that's different appears to be that the change from:   `const description = props.edit.description.replace(/"/g, '')` to  `const cleanDescription = props.edit.description.replace(/^"|"$/g, '')` fixes a bug where quotations marks were stripped from inside two edits which is a plus:

Screenshot 2026-08-19 at 12 25 02 PM
Screenshot 2026-08-19 at 12 24 42 PM

I could imagine making up some fun, less janky tests to check over things in a future PR?

contolini (approved):
🥮

PR #2820, opened 2026-07-31
test: update cypress tests to match new institution names and data

contolini:
2023-2025 institution data fields were updated this morning from their relevant transmittal sheets, causing our cypress tests to need new selectors.

## Testing

1. Run cypress against ffiec.cfpb.gov and it should be green.


billhimmelsbach (approved):
Checked out the tests and confirmed they are passing now, so I'm approving this one and changing the cron jobs to the new image. 🚀

PR #2837, opened 2026-09-10
fix(filing): remove duplicate ULI column from Q646 edit table

contolini:
For a small number of edits, our API returns the loan's ULI as the first field in the fetched response. This results in tables having two ULI columns since the frontend [adds a ULI column](https://github.com/cfpb/hmda-frontend/blob/1c084940043cabd735dbd739f3ab428b32b42d4b/src/filing/submission/edits/Table.jsx#L34-L41) as the first column of every edit table. A `suppressULI` function was added [back in the day](https://github.com/cfpb/hmda-frontend/pull/368) to fix this situation.

This PR added Q646 to the list of edits that need their ULI suppressed.

## Testing

1. Tests should pass.
2. Uploading the `Step2_` file from user testing should show only one ULI column in the quality edits table.


billhimmelsbach (approved):
Looks good!

PR #2840, opened 2026-09-10
fix(profile): change save endpoint to never point at beta

billhimmelsbach:
# Don't merge, testing it out on staging now

Users were reporting that they were unable to change user profile information on beta such as name, associated institutions, etc. Looks like it's an old bug where profile changes were being posted to an auth endpoint with the same origin as the environment, even though beta doesn't have its own auth service.

## Changes

- strip `.beta` or `-beta` from the profile endpoint if `isBeta` is true

## Testing

1. How does it look on staging? (both staging and staging-beta)

## Screenshots

**Staging Beta**

**Staging**


billhimmelsbach:
Don't need this now that we have duplicated the auth service.

PR #2845, opened 2026-09-21
fix(filing): horizontally scroll edits table (not its pagination or caption)

contolini:
Moves edit tables' `<caption>` elements outside their `<table>` elements into a div above the tables. This lets the tables horizontally scroll while their titles and descriptions remain sticky. Pagination also remains sticky underneath the tables. Accessible table names and descriptions are preserved using `aria-labelledby` and `aria-describedby` attributes.

See [#5432](https://github.local/HMDA-Operations/hmda-devops/issues/5432).

## Changes

- Refactored edits tables.

## Testing

1. Tests should pass.
2. To test locally, you can use the `Step2_` file from user testing to generate a paginated edit table.

## Screenshots

| before | after |
|--------|-------|
| scroll-before | scroll-after |


billhimmelsbach (approved):
This change is small and great, and I appreciate keeping things accessible after moving the caption around. Would you mind resolving the git conflict after the [markdown changes](https://github.com/cfpb/hmda-frontend/pull/2815), then we're good to go?

